PR #14, opened 2021-01-29
support multi-tags

sdelamo:
@ilopmar what do you think?


ilopmar:
Awesome! The only thing I'd change is using another separator for the tags, like the `|`:

```
source:BookCatalogueClient[app=bookrecommendation,tags=packageandimports|harcoded|clazz]
```

ilopmar (approved):

PR #55, opened 2021-01-29
Fix single and multi-tags

ilopmar:
This PR fixes two things:

- Multi-tags: Use `tokenize` instead of `split` because the latter receives a regex you using `|` doesn't work. It was splitting all the characters in the tag.
- Single tag: When there wasn't a tag defined the previous code added a `tag=` to every include.

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
+++ b/buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
@@ -201,9 +201,15 @@ class GuideAsciidocGenerator {
     static List<String> extractTags(String line, String tagSeparator = '|') {
         String attributeValue = extractFromParametersLine(line, 'tags')
         if (attributeValue) {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
```diff
@@ -201,9 +201,15 @@ class GuideAsciidocGenerator {
     static List<String> extractTags(String line, String tagSeparator = '|') {
         String attributeValue = extractFromParametersLine(line, 'tags')
         if (attributeValue) {
-            return attributeValue?.split(tagSeparator) as List<String>
+            return attributeValue.tokenize(tagSeparator)
+        }
+
+        String singleTag = extractTagName(line)
```
sdelamo: I am going to replace

```

         String singleTag = extractTagName(line)
         if (singleTag.isEmpty()) {
             return []
         } else {
             return [singleTag]
         }
```

with: 

`[extractTagName(line)].findAll { it }`

PR #91, opened 2021-01-30
Fix test.sh script and run tests in GitHub actions

### LINKED ISSUE #4: Setup github action to run tests
There should be a github actions which does `./gradlew build` and then `cd code;./test.sh`

### LINKED ISSUE #9: Modify test.sh
@ilopmar 
> I've also added support to read skipGradleTests and skipMavenTests from metadata and default them to false if they are not present. What is missing is modifying the logic to generate the appropriate tesh.sh for multi-projects and take those flags into account.

see: https://github.com/sdelamo/micronaut-guides-poc/pull/8

sdelamo:
Close: https://github.com/micronaut-projects/micronaut-guides-poc/issues/4
Close: https://github.com/micronaut-projects/micronaut-guides-poc/issues/9

This PR: 

- Fixes the test script
- Run the tests in Github actions
- Run them in a JDK matrix by generating projects for different JDKs
- Address an issue with a Groovy Guide which was using @Fallback but the @Client were missing @Recoverable

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,10 +67,11 @@ task generateSampleProjects {
     ext {
         metadataConfigName = 'metadata.json'
         output = "$buildDir/code"
+        guides = "${projectDir}/guides"
     }
     doLast {
         GuideProjectGenerator generator = new GuideProjectGenerator()
-        generator.generate(new File("${projectDir}/guides"), new File(output), metadataConfigName, true, new File("${projectDir}/src/docs/asciidoc"))
+        generator.generate(new File(guides), new File(output), metadataConfigName, true, new File("${projectDir}/src/docs/asciidoc"))
         generator.close()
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -81,13 +82,13 @@ task generateTestScript {
     description 'Generates a test.sh to runs every guide test at build/code'
     ext {
         excludeDirs = ['images']
-        stopIfFailure = false
+        stopIfFailure = true
```
ilopmar: Why have you changed this? I think it makes more sense to run always all the tests of all the guides.

sdelamo: I changed it accidentally. I was testing locally and I forgot to change it back. I have reverted the change.

ilopmar (approved):

PR #101, opened 2021-02-09
New workflow for testing Micronaut Snapshot

ilopmar:
Finally have it working with the approach of modifying the `version.txt` file with the same bash script we use in our Github Actions for setting the next snapshot. Then I just run the build and the generated `test.sh`.

The job runs properly: https://github.com/micronaut-projects/micronaut-guides-poc/runs/1864357635?check_suite_focus=true#step:5:1

I've modified the job to set a cron on Sundays at 05:00 and force the push.

---
@sdelamo I've also tried a different approach. Modify the version file in as part of a gradle task. The problem with this approach is that we can't start with 2.3.1 and in the same "run" modify the file. Gradle won't see the changes. So this is basically a two step task: first modify the version and then run the build. That's why I decided to go with the script approach because it is easier.
I have a couple of local branches with the things I've tried in case you want to take a look at them.

### DIFF
--- a/increment_version.sh
+++ b/increment_version.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# Increment a version string using Semantic Versioning (SemVer) terminology.
+# Source: https://github.com/fmahnke/shell-semver
+
+# Parse command line options.
+
+while getopts ":Mmp" Option
+do
+  case $Option in
+    M ) major=true;;
+    m ) minor=true;;
+    p ) patch=true;;
+  esac
+done
+
+shift $(($OPTIND - 1))
+
+version=`cat version.txt`
+
+# Build array from version string.
+
+a=( ${version//./ } )
+
+# Increment version numbers as requested.
+
+if [ ! -z $major ]
+then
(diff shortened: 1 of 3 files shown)


increment_version.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+# Increment a version string using Semantic Versioning (SemVer) terminology.
+# Source: https://github.com/fmahnke/shell-semver
```
sdelamo, line 4: bash for the win!

sdelamo (approved):
This looks good for the patch release. We will need to think how to test the next minor version. I expect we will face more breaking changes there.

ilopmar:
@sdelamo do you want to test also the next minor? Then I think we need another workflow for it or test one version after the other. 
The script allows the param `-m` to calculate the next minor and `-M` for the next major.

sdelamo:
Yes, I think testing the next minor will be valuable since the breaking changes may be there.

ilopmar:
Submitted PR #106

PR #119, opened 2021-02-17
Add new macro to group dependencies in same block

ilopmar:
This PRs adds the possibility to group a list of dependencies in the same group. This is used in the Zipkin guide (and probably in others):

The following code:

```
:dependencies:

dependency:micronaut-tracing[groupId=io.opentracing.brave]
dependency:brave-instrumentation-http[groupId=io.zipkin.brave,scope=runtimeOnly]
dependency:zipkin-reporter[groupId=io.zipkin.reporter,scope=runtimeOnly]

:dependencies:
```

Generates all the dependencies inside the same block (right screenshot) instead of in separate blocks (left screenshot):

DeepinScreenshot_select-area_20210217132439


sdelamo (approved):
@ilopmar nice!

PR #145, opened 2021-03-10
New macro for tests without logic to detect language/file extension

ilopmar:
For some guides like Security Cookie we use Geb for tests and we also include snippets of those tests in the guide. 
The problem with our current approach is that we include a test file with `test:HomeTest[]` and depending on the language and the test framework we look for the right file in `src/test/{java,groovy,kotlin}` directories, with the right file extension, and even for Spock, we look for `HomeSpec.groovy` instead of `HomeTest.groovy`. This makes really easy to write multi-language guides.

But now, we use Geb for testing the three languages (so all the test code is Groovy), so our previous macro doesn't work anymore. There isn't an easy way to know that when doing `test:HomePage[]` we really want to include the file `src/test/groovy/.../HomePage.groovy` independently of the language of the guide.

I've created a new macro `rawTest:` that just includes the file without guessing anything and only based on the test framework. Even though the code seems similar to what we already have I've decided to create a new method to avoid adding more `if` in the other one (and to break things). 
If we want we can refactor in the future.

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
+++ b/buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
@@ -7,6 +7,11 @@ import io.micronaut.starter.api.TestFramework
 import java.nio.file.Path
 import java.nio.file.Paths
 
+import static io.micronaut.starter.api.TestFramework.JUNIT
+import static io.micronaut.starter.api.TestFramework.KOTEST
+import static io.micronaut.starter.api.TestFramework.KOTLINTEST
+import static io.micronaut.starter.api.TestFramework.SPOCK
+
 @CompileStatic
 class GuideAsciidocGenerator {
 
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
```diff
@@ -252,4 +288,32 @@ class GuideAsciidocGenerator {
...
+                return 'groovy'
+            case KOTEST:
+            case KOTLINTEST:
+                return 'kt'
+            default:
+                return '@languageextension@'
+        }
+    }
+
+    private static String resolveLangFolder(TestFramework testFramework) {
```
sdelamo: I think we already have methods for this in starter. see: `Language::getTestDir`

buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
```diff
@@ -252,4 +288,32 @@ class GuideAsciidocGenerator {
                 return ''
         }
     }
+
+    private static String resolveTestExtension(TestFramework testFramework) {
+        switch (testFramework) {
```
sdelamo: I think we have methods for this in starter: `Language:getExtension`

buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
ilopmar: Done!

buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
ilopmar: Done

sdelamo (approved):

PR #149, opened 2021-03-10
Add Download Excel file guide

### LINKED ISSUE #45: Migrate Download an Excel file in Micronaut App

ilopmar:
Closes #45

This can't be merge because:
- We don't have a way to contribute to Gradle plugins or additional Gradle configuration in features created in the guides. Also the same for Maven.
- The original guide only used Gradle and it used a Gradle plugin to download and configure Webdriver. It seems there are another one for Maven but we need to figure it out how to use it.
- [common-webdriverbinaries.adoc](https://github.com/micronaut-projects/micronaut-guides-poc/pull/149/files#diff-c3b3cea8f6aaabf04c849f3df360daf596473edfdc8426392603f2c901a795d6) included in this PR needs to be adapted to Maven.
- The same for running the tests selecting Chrome with Maven. See [here](https://github.com/micronaut-projects/micronaut-guides-poc/pull/149/files#diff-49a8f965b7cf3953ef21216165db168be2533bf4ea1aa7d13b36363c82d12ef5R124)


ilopmar:
Ready for review

buildSrc/build.gradle
```diff
@@ -1,6 +1,7 @@
 plugins {
     id "java-gradle-plugin"
     id 'groovy'
+    id 'nu.studer.rocker' version '3.0.2'
```
sdelamo: I don't think this is the correct rocker plugin. We typically use: 

https://plugins.gradle.org/plugin/com.fizzed.rocker

sdelamo (changes requested):

buildSrc/build.gradle
ilopmar: This plugins works with Gradle 7 and uses the same `com.fizzed.rocker` core library under the hood.

sdelamo: How would you feel if we remove the plugin and just write without a rocker template? I don't like adding a non official plugin. 

```
webdriverBinaries {
     chromedriver {
         // Get latest for your OS and Chrome from
         // https://github.com/webdriverextensions/webdriverextensions-maven-plugin-repository/blob/master/repository-3.0.json
         version = '88.0.4324.96'
     }
 }

 tasks.withType(Test) {
     systemProperty "geb.env", System.getProperty('geb.env')
     systemProperty "download.folder", System.getProperty('download.folder')
 }
 ```

ilopmar: "Official" is a strong word :stuck_out_tongue:. I think this plugin is better than the other one and it works with Gradle 7 (remember that it uses the same library under the hood, but in this case it is just a Gradle plugin).

Having said that, for the guides I don't think there is any difference in using this plugin or rendering the template as a String. At least, for this case in which we don't need to interpolate any value. I can do the change and remove the plugin but if we need more powerful templates in the future in the guides, I'll bring it back.

ilopmar: Done

ilopmar:
@sdelamo this should be ok to final review and merge

sdelamo (approved):

sdelamo:
@ilopmar I have removed the Geb test which required Chrome and added a test which downloads the file and uses the byte array. That way the test will be run on CI.

I have add Junit test for Java and create Groovy / spock version of the guide.

ilopmar:
Excellent! :100:

PR #175, opened 2021-04-06
Add micronaut-azure-cloud guide

ZacharyKlein:
This guide apparently depends on a Maven plugin, which means it doesn't work for Gradle projects (as currently written).  There does appear to be a Grade plugin available, however it looks to be unofficial, and the repo hasn't been updated in over a year. https://github.com/lenala/azure-gradle-plugins


ilopmar (dismissed):
Please remove the Maven wrapper files.

ilopmar:
It is possible to configure the guide to use only Maven. Add `"buildTools": ["maven"]` to the `metadata.json` file.

guides/micronaut-azure-cloud/micronaut-azure-cloud.adoc
```diff
@@ -0,0 +1,202 @@
...
+== How to Complete
+
+To get started do the following:
+
+Download and unzip the source or clone the following GitHub repository:
+
+[source, bash]
+----
+git clone https://github.com/micronaut-guides/micronaut-azure-cloud.git
+----
```
ilopmar: This shouldn't be necessary. The users either download the .zip of the full project or create the app step by step.

guides/micronaut-azure-cloud/micronaut-azure-cloud.adoc
```diff
@@ -0,0 +1,202 @@
...
+
+.pom.xml
+[source,xml]
+----
+      <plugin>
+        <groupId>com.microsoft.azure</groupId>
+        <artifactId>azure-webapp-maven-plugin</artifactId>
+        <version>1.6.0</version>
+      </plugin>
+----
```
ilopmar: Please use our `dependency:xxxxx` macron

guides/micronaut-azure-cloud/micronaut-azure-cloud.adoc
```diff
@@ -0,0 +1,202 @@
...
+[INFO] ------------------------------------------------------------------------
+[INFO] BUILD SUCCESS
+[INFO] ------------------------------------------------------------------------
+----
+
+== Sample Controller
+
+Create a controller which we can invoke once the app is deployed.
+
+source:HelloController.java[]
```
ilopmar: This fails to find the file. It should be `source:HelloController[]`

ilopmar (changes requested):
Additional comments:
- We should follow the steps in the guide to make sure this still works because the original guide is quite old (June 2019)
- We should probably update this to use Java 11 which I think it's available in Azure.

ilopmar (approved):
Good work @ZacharyKlein :tada:

PR #184, opened 2021-04-13
Guides should have Micronaut Website branding

sdelamo:
It probably needs still some adjustments but I think it is in the direction we are looking for. Let me know if you think it is fine to merge it. 

See: https://github.com/micronaut-projects/micronaut-guides-poc/issues/171

### DIFF
--- a/assets/template.html
+++ b/assets/template.html
@@ -1,0 +1,223 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
+    <meta name='robots' content='index, follow, max-image-preview:large, max-snippet:-1, max-video-preview:-1'/>
+    <title>@title@</title>
+    <link rel="canonical" href="@canonical@"/>
+
+    <link rel='stylesheet' id='wp-block-library-css' href='https://micronaut.io/wp-includes/css/dist/block-library/style.min.css' type='text/css' media='all'/>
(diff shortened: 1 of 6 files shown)


assets/template.html
```diff
@@ -0,0 +1,223 @@
...
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
+    <meta name='robots' content='index, follow, max-image-preview:large, max-snippet:-1, max-video-preview:-1'/>
+    <title>@title@</title>
+    <link rel="canonical" href="@canonical@"/>
+
+    <link rel='stylesheet' id='wp-block-library-css' href='https://micronaut.io/wp-includes/css/dist/block-library/style.min.css' type='text/css' media='all'/>
+    <link rel='stylesheet' id='fonts-fira-sans-css' href='//fonts.googleapis.com/css2?family=Fira+Sans%3Awght%40300%3B500&#038;display=swap&#038;ver=5.7' type='text/css' media='all'/>
```
ilopmar: Missing `https:`

ilopmar (changes requested):
The output is not 100% like the design but looks really good and promising!
We need to adjust a few things like setting in bold the subtitle, get the icons for the asciidoctor `NOTE`, `WARNING`,... and probably some other minor adjustments, but I really like how the guides look like :100: 

There are some CORS errors getting the fonts directly from Micronaut site.

image

assets/template.html
sdelamo: I added the https. I don't see the cors error. Maybe that got rid of the error.

sdelamo: > We need to adjust a few things like setting in bold the subtitle, get the icons for the asciidoctor NOTE, WARNING,... 

I added fontawesome which powers that.

sdelamo:
@ilopmar I am going to merge this to progress in the design.

PR #190, opened 2021-04-21
Add language/builds to categories and fix category names

ilopmar:


buildSrc/src/main/groovy/io/micronaut/guides/IndexGenerator.groovy
```diff
@@ -243,8 +243,8 @@ class IndexGenerator {
                 title: guide.title,
                 intro: guide.intro,
                 authors: guide.authors,
-                tags: guide.tags,
-                category: guide.category,
+                tags: guide.tags + guide.languages.collect { it.toString() } + guide.buildTools.collect { it.toString() },
```
sdelamo, line 246: @ilopmar if you were cool, you will write this as `guides.languages*.toString()`

ilopmar: :see_no_evil: I used to be cool but it's been so long since I wrote Groovy (apart from tests)

PR #194, opened 2021-04-28
Micronaut multitenancy propagation

### LINKED ISSUE #181: Migrate "Micronaut Multitenancy Propagation" guide
https://guides.micronaut.io/micronaut-multitenancy-propagation/guide/index.html

ZacharyKlein:
- Resolves https://github.com/micronaut-projects/micronaut-guides-poc/issues/181
- Migrated demo applications to Groovy as the new infrastructure doesn't support mixed-language guides
- Added `Multitenancy` and `MultitenancyGorm` custom features (`MutitenancyGorm` is language specific to Groovy).
- Revised guide content to match new code
- Verified that the applications work as expected and tests are passing


ilopmar (changes requested):
I just pushed some minor improvements but everything looks good, although the tests are failing.

ZacharyKlein:
@ilopmar Is there a way to view test results from a Github action? I cannot reproduce the failures locally, even with Java 15.  I don't see any error messages in the logs either, so I'm not sure what test is failing or what to do differently.

ZacharyKlein:
@sdelamo Can you look at this build? I cannot reproduce the test failures. I've tried multiple Java versions, running the entire test suite (all guides), running the tests from the generated projects directly, and running the application and testing the same flow manually.  I'm not sure what else might be at play, other than perhaps the data in the Bootstrap class isn't being created before the test runs?

ilopmar:
To troubleshoot this better you can add the following to the workflow to save the reports: https://github.com/micronaut-projects/micronaut-guides-poc/blob/issue-202/.github/workflows/gradle.yml#L33-L38

Also add/modify a custom `logback.xml` with more debug information for the HTTP Client. This way you will see exactly why the tests fail. I did the same to figure out why tests with http client were failing in the new micronaut-http-client guide.

P.S: The change in the workflows are part of my PR #206, so once it's merged you can merge master into this branch (or in the meantime you can just apply the workflow changes manually)

sdelamo:
I have found the issue. I think it was a test race condition. https://github.com/micronaut-projects/micronaut-guides-poc/pull/194/commits/6a1a9f5e7bef72ad0ca4f0304db42566e767c667

buildSrc/src/main/java/io/micronaut/guides/feature/Multitenancy.java
```diff
@@ -0,0 +1,33 @@
+package io.micronaut.guides.feature;
```
sdelamo: There is already a feature. It is named `multi-tenancy`

sdelamo (changes requested):

sdelamo (changes requested):
The image is not related to the guides content. 

See the image used in the old guide: 

https://guides.micronaut.io/micronaut-multitenancy-propagation/img/tokenpropagation.svg

PR #219, opened 2021-05-06
Create CNAME

sdelamo:
See: https://github.com/micronaut-projects/micronaut-guides-poc/issues/218

See: https://github.com/micronaut-projects/micronaut-guides/blob/gh-pages/CNAME

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+guides.micronaut.io


sdelamo:
This is the deletion: https://github.com/micronaut-projects/micronaut-guides/pull/11

I think merging both @JasonTypesCodes should do the magic.

ilopmar:
I think will should leave this PR and the other one in "Draft" until we are ready to do the switch.

PR #225, opened 2021-05-07
Resolve version dependencies used in asciidoctor

ilopmar:
This PR allows to extract external dependencies versions from the pom.xml included in the project and use them in the asciidoctor guide.

So, given the following in `pom.xml`:
```xml
...
        <dependency>
            <groupId>org.gebish</groupId>
            <artifactId>geb-spock</artifactId>
            <version>4.1</version>
        </dependency>
...
```

We can write the following in the guide:

```
dependency:geb-spock[groupId=org.gebish,scope=testImplementation,version=@geb-spockVersion@]
```

This is the `@` + `artifactId` + `Version` + `@` and it will be replaced with the value defined in the pom (that it's automatically updated via dependabot).


ilopmar:
@sdelamo ready to review :)

sdelamo (approved):

PR #256, opened 2021-05-25
Add GraphQL guide

ilopmar:


sdelamo (approved):
I've done some changes to make the Kotlin code more idiomatic. 

I have also changed so that graphiQL is only enabled for the default environment which I have set to dev.

ilopmar:
Thanks for the improvements on the Kotlin code. Don't blame me but IntelliJ, it did the conversion automatically :see_no_evil: 

Regarding the default environment and GraphiQL, I don't agree with the change. It is dangerous keeping it enabled by default, which is what happens now when users don't select any environment. So, if you just package the app and deploy it you will have that endpoint enable in production.

sdelamo:
> Regarding the default environment and GraphiQL, I don't agree with the change. It is dangerous keeping it enabled by default, which is what happens now when users don't select any environment. So, if you just package the app and deploy it you will have that endpoint enable in production.

Environment deduction is enabled by default. Because of that, when you deploy your application, 99% of the time another environment ('ec2', 'azure', 'k8s') will be detected and the default environment will not be loaded.  And thus, graphiql will not be loaded in production. 

I think it is safer to tell the user to create a an `application-dev.yml` file and setup `dev` as a default environment than to enable `graphiql` in `application.yml` and ask him to remember to delete that configuration.

ilopmar:
You are assuming that they are going to deploy to some cloud, k8s,... what I they just copy the jar to some machine and just start it?

sdelamo:
In the Micronaut documentation, we suggest to use the default environment as a development environment: 

https://docs.micronaut.io/latest/guide/#_default_environment 

If you think we should not suggest the default environment as a development environment then I would suggest that you create a PR against that part of the docs. 

It seems you feel strongly about this, I encapsulate the change in a single commit. Revert it and merge the PR.

ilopmar:
Thank you Sergio!

PR #257, opened 2021-05-27
s3 event lambda guide

sdelamo:
I've added in this PR the ability to delete some files of the generated apps. @ilopmar let me know if you want me to extract that feature to a separate PR.


ilopmar (approved):
Really cool feature to remove generated files! :)

Btw, this PR includes two guides `micronaut-aws-lambda-s3-event` and `micronaut-aws-lambda-eventbridge-event`. Which is fine, you don't need to split them in separate PRs if you don't want.

sdelamo:
> Btw, this PR includes two guides micronaut-aws-lambda-s3-event and micronaut-aws-lambda-eventbridge-event. Which is fine, you don't need to split them in separate PRs if you don't want.

Yes, both guides needed the delete files features.

PR #258, opened 2021-05-28
Micronaut Data JDBC Guide

sdelamo:
I have written a Micronaut Data JDBC Guide which is a copy of JPA Hibernate guide


guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreRepository.java
```diff
@@ -0,0 +1,67 @@
...
+
+    List<Genre> findAllOrderById();
+
+    List<Genre> findAllOrderByIdDesc();
+
+    List<Genre> findAllOrderByName();
+
+    List<Genre> findAllOrderByNameDesc();
+
+    default List<Genre> findAll(@NotNull SortingAndOrderArguments args) {
```
dstepanov: Can this be rewritten to use `Pageable`?

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/ApplicationConfiguration.java
```diff
@@ -0,0 +1,8 @@
+package example.micronaut;
+
+import javax.validation.constraints.NotNull;
+
+public interface ApplicationConfiguration {
+
+    @NotNull Integer getMax();
+}
```
graemerocher, line 8: Why the need for an interface and an implementation? Isn't it simpler to just write:

```java
@ConfigurationProperties("application")
public interface ApplicationConfiguration {   
    @NotNull Integer getMax();
}
```

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreController.java
```diff
@@ -0,0 +1,84 @@
...
+    public GenreController(GenreRepository genreRepository) { // <3>
+        this.genreRepository = genreRepository;
+    }
+
+    @Get("/{id}") // <4>
+    public Genre show(Long id) {
+        return genreRepository
+                .findById(id)
+                .orElse(null); // <5>
+    }
```
graemerocher: Isn't it better to just return `Optional<Genre>`?

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreController.java
```diff
@@ -0,0 +1,84 @@
...
+    public HttpResponse update(@Body @Valid GenreUpdateCommand command) { // <7>
+        int numberOfEntitiesUpdated = genreRepository.update(command.getId(), command.getName());
+
+        return HttpResponse
+                .noContent()
+                .header(HttpHeaders.LOCATION, location(command.getId()).getPath()); // <8>
+    }
+
+    @Get(value = "/list{?args*}") // <9>
+    public List<Genre> list(@Valid SortingAndOrderArguments args) {
```
graemerocher: Should use `Pageable`

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreController.java
```diff
@@ -0,0 +1,84 @@
...
+        } catch(DataAccessException e) {
+            return HttpResponse.noContent();
+        }
+    }
+
+    @Delete("/{id}") // <12>
+    public HttpResponse delete(Long id) {
+        genreRepository.deleteById(id);
+        return HttpResponse.noContent();
+    }
```
graemerocher: This is cleaner as:

```java

@Delete("/{id}")
@Status(NO_CONTENT)
void delete(long id) {
    genreRepository.deleteById(id);
}
```

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreRepository.java
graemerocher: Should in fact extend from `PageableRepository` and then you get the method for free

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/domain/Book.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+import javax.validation.constraints.NotNull;
+
+@Entity
+@Table(name = "book")
```
graemerocher, line 13: Why the use of javax annotations? Seems unnecessary

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreSaveCommand.java
```diff
@@ -0,0 +1,27 @@
...
+
+import io.micronaut.core.annotation.Introspected;
+
+import javax.validation.constraints.NotBlank;
+
+@Introspected // <1>
+public class GenreSaveCommand {
+
+    @NotBlank
+    private String name;
```
graemerocher: Should be final

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreSaveCommand.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public GenreSaveCommand() {
+    }
```
graemerocher: Why the extra constructor?

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreSaveCommand.java
```diff
@@ -0,0 +1,27 @@
...
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
```
graemerocher: Why the setter if it is never used?

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreUpdateCommand.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+
+@Introspected
+public class GenreUpdateCommand {
+    @NotNull
+    private Long id;
+
+    @NotBlank
+    private String name;
```
graemerocher: Should both be final

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreUpdateCommand.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public GenreUpdateCommand() {
+    }
```
graemerocher: Why the extra constructor?

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreUpdateCommand.java
```diff
@@ -0,0 +1,39 @@
...
+        this.name = name;
+    }
+
+    public Long getId() {
+        return id;
+    }
+
+    public void setId(Long id) {
+        this.id = id;
+    }
```
graemerocher: Delete unnecessary setter

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreUpdateCommand.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
```
graemerocher: Delete unnecessary setter

graemerocher (changes requested):
Nice start, I've provided some feedback. Once addressed will send a subsequent PR as well.

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
```diff
@@ -0,0 +1,237 @@
...
+resource:application.yml[tag=datasource]
+
+NOTE: This way of defining the datasource properties means that we can externalize the configuration, for example for
+production environment, and also provide a default value for development. If the environment variables are not defined
+Micronaut will use the default values. +
+Also keep in mind that it is necessary to escape the `:` in the connection url using back ticks ```.
+
+=== DB Schema
+
+Now that Mybatis is setup we need a way to create the database schema. For that we will use
```
graemerocher: Mybatis?

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
```diff
@@ -0,0 +1,237 @@
...
+source:domain/Book[]
+
+
+=== Application Configuration
+
+Create an interface to encapsulate the application configuration settings:
+
+source:ApplicationConfiguration[]
+
+In Micronaut, like Spring Boot and Grails, you can create type safe configuration by creating classes that are annotated
```
graemerocher: No need to mention spring and grails imo

graemerocher (changes requested):

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreRepository.java
sdelamo: I've extended from `PageableRepository` and improved the repository section copy.

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/ApplicationConfiguration.java
sdelamo: I prefer pure interfaces decoupled from `@ConfigurationProperties` implementations.

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
sdelamo: This is a copy past from the mybatis guide where we introduced the flyway section. I have fixed it.

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
sdelamo: removed

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreUpdateCommand.java
sdelamo: deleted

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreUpdateCommand.java
sdelamo: deleted

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreUpdateCommand.java
sdelamo: deleted

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreUpdateCommand.java
sdelamo: done

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreSaveCommand.java
sdelamo: deleted

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreController.java
sdelamo: I rewrote it as you suggest

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreController.java
sdelamo: changed

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/ApplicationConfiguration.java
graemerocher: Just seems to add more verbosity to the example IMO

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreController.java
sdelamo: I've changed to `Pageable` as a controller method argument. 

A couple of changes, I had to do:

- When binding Pageable as a method argument sort is bound as `sort=name,desc` in our examples we had something like `sort=name&order=desc'.

We had a test for an invalid order. 

```java
@Test
    public void supplyAnInvalidOrderTriggersValidationFailure() {
        HttpClientResponseException thrown = assertThrows(HttpClientResponseException.class, () -> {
            client.toBlocking().exchange(HttpRequest.GET("/genres/list?order=foo"));
        });

        assertNotNull(thrown.getResponse());
        assertEquals(HttpStatus.BAD_REQUEST, thrown.getStatus());
    }
```

Rewriting it as follows does not thrown any validation exception but binds to the default sorting which is `ASC`


```java
@Test
    public void supplyAnInvalidOrderTriggersValidationFailure() {
        HttpClientResponseException thrown = assertThrows(HttpClientResponseException.class, () -> {
            client.toBlocking().exchange(HttpRequest.GET("/genres/list?order=name,foo"));
        });

        assertNotNull(thrown.getResponse());
        assertEquals(HttpStatus.BAD_REQUEST, thrown.getStatus());
    }
```

I have removed the test.

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/domain/Book.java
sdelamo: This guide is a port of [Access a database with mybatis](https://guides.micronaut.io/latest/micronaut-data-access-mybatis.html) and [access a database with JPA and Hibernate](https://guides.micronaut.io/latest/micronaut-jpa-hibernate.html).  I think it is a good example of how Micronaut Data JDBC simplifies development since this guide basically deletes code because the repository is implemented by Micronaut Data. 

`Genre`, the entity class, is identical in those guides and in this Micronaut Data JDBC guide. I think it is valuable to show that you can use the JPA annotations you are used to.

sdelamo:
> Nice start, I've provided some feedback. Once addressed will send a subsequent PR as well.

I think I have addressed every thing. I have commented about the jaxax.persistence usage  here: https://github.com/micronaut-projects/micronaut-guides/pull/258#discussion_r642111291 and about the binding of Pageable https://github.com/micronaut-projects/micronaut-guides/pull/258#discussion_r642242514

@graemerocher please provide your subsequent PR.

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
```diff
@@ -0,0 +1,254 @@
...
+
+Add the following dependencies:
+
+:dependencies:
+
+
+implementation("io.micronaut:micronaut-http-client")
+implementation("io.micronaut:micronaut-runtime")
+implementation("io.micronaut.data:micronaut-data-hibernate-jpa")
+implementation("io.micronaut.flyway:micronaut-flyway")
```
ilopmar: I think these should be removed.

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
```diff
@@ -0,0 +1,254 @@
...
+source:ApplicationConfiguration[]
+
+In Micronaut, you can create type safe configuration by creating classes that are annotated
+with https://docs.micronaut.io/latest/guide/index.html#configurationProperties[@ConfigurationProperties].
+
+Create a `ApplicationConfigurationProperties` class:
+
+source:ApplicationConfigurationProperties[]
+
+<1> @ConfigurationProperties` annotation takes a configuration prefix.
```
ilopmar: Missing a back tick

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreController.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Get("/{id}") // <4>
+    public Optional<Genre> show(Long id) {
+        return genreRepository
+                .findById(id); // <5>
+    }
+
+    @Put // <6>
+    public HttpResponse update(@Body @Valid GenreUpdateCommand command) { // <7>
+        int numberOfEntitiesUpdated = genreRepository.update(command.getId(), command.getName());
```
ilopmar: The value of `numberOfEntitiesUpdated` is not used. Maybe you can remove the variable and call the method in the repository.

ilopmar (changes requested):
Besides my minor comments, I'll add back the GraalVM section because we know that it works properly.

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/domain/Book.java
graemerocher: The guides are meant to show how we believe developers should write applications. I don't think it is the case that we recommend developers use JPA annotations. It is just another dependency you have to add which is unnecessary. If we want to call it out by pointing to the Micronaut Data documentation that seems reasonable,  but I don't see why we should be recommending JPA annotations by default. In particular it just forces the developer to add another dependency. If it was our recommending when you created an app with Micronaut Data JDBC in Launch the JPA dependency should be there as well which it is not.

ilopmar: I see this the same way as Sergio. I think developers are used to use the JPA annotations and learn ours annotations is an additional extra step that some of them probably don't want to do. Specially when the JPA annotations work and they have better support in IDEs.

graemerocher: Ok so in that case why aren't we creating the project in launch with the JPA annotations included?

graemerocher: Also I must point out that the point of Micronaut Data JDBC is not to accommodate what developers are used to. The example in fact is using mutable entities and getters/setters

Micronaut Data JDBC's design is to encourage the use of immutable entities and mapping JDBC results to entities. It is not an ORM tool but a simple data mapper, therefore we should be showing patterns that demonstrate how it is designed to be used, not just porting a JPA application and using the same patterns.

graemerocher: In addition if we are going to use the JPA annotations we will need to point out in the guide that with JDBC only a subset of the JPA annotations are supported and that since Data JDBC is not a full ORM developers cannot expect that every annotation works. We should point to a reference documentation which describes which annotations are in fact supported otherwise this will lead to confusion for readers of the guide.

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
sdelamo: done

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/GenreController.java
sdelamo: I've removed the unused variable

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
sdelamo: done.

sdelamo:
@ilopmar 
> Besides my minor comments, I'll add back the GraalVM section because we know that it works properly.

I've added the GraalVM section.

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/SortingAndOrderArguments.java
```diff
@@ -0,0 +1,65 @@
...
+import io.micronaut.core.annotation.Introspected;
+
+import io.micronaut.core.annotation.Nullable;
+import javax.validation.constraints.Pattern;
+import javax.validation.constraints.Positive;
+import javax.validation.constraints.PositiveOrZero;
+import java.util.Optional;
+
+@Introspected
+public class SortingAndOrderArguments {
```
graemerocher: This can be deleted now can it not?

graemerocher (changes requested):

guides/micronaut-data-jdbc-repository/java/src/main/java/example/micronaut/SortingAndOrderArguments.java
sdelamo: deleted

ilopmar (approved):
We should probably discuss the JPA annotations vs Micronaut Data tomorrow.

Besides that, everything looks good.

graemerocher:
What happened to discussing the JPA annotations?

I gotta say I don't know why you guys asked for feedback from the folks who wrote Micronaut Data if you aren't going to take it onboard 🤷‍♂️

ilopmar:
@sdelamo was too fast with the merge...

PR #265, opened 2021-06-01
mn cli option should be `features`

burtbeckwith:


ilopmar (approved):

ilopmar:
Good catch!

sdelamo:
> Good catch!

Indeed. Great to have your 👀 on this project @burtbeckwith !

PR #272, opened 2021-06-02
feat: use micronaut data annotations

sdelamo:


guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
```diff
@@ -80,6 +76,7 @@ The previous domain has a `OneToMany` relationship with the domain `Book`.
 
 source:domain/Book[]
 
+TIP: You could use JPA annotations instead by including the following `compileOnly` scoped dependency `jakarta.persistence:jakarta.persistence-api`.
```
graemerocher: Add a link to https://micronaut-projects.github.io/micronaut-data/latest/guide/#sqlAnnotations which describes the supported annotations and which ones are supported

graemerocher (commented):
I have some questions since the example is confusing to me.

* The guide creates ApplicationConfiguration and an implementation configuration properties but this is not used anywhere, so why is this necessary for the guide?
* A Book entity is created but then not used anywhere in the example, nor is association handling demonstrated or tested anywhere. Why is it there?
* Why create command objects like GenreSaveCommand which only have a single property? Is there some functional reason for this?
Once I have answers to these questions I will send a PR to the branch you created. Thanks!

graemerocher:
Any feedback?

sdelamo:
working on your suggestions now.

sdelamo:
@graemerocher :

> * The guide creates ApplicationConfiguration and an implementation configuration properties but this is not used anywhere, so why is this necessary for the guide?

In the [Micronaut SQL Hibernate JPA guide](http://localhost/guides/micronaut-data-jdbc-repository-gradle-java.html#data-source-configuration) we use `application.max` to show the user how we use configuration to limit the maximum number of results. E.g: 

`query.setMaxResults(args.getMax().orElseGet(applicationConfiguration::getMax));`

In this guide, it makes more sense to tell the user that he can use `micronaut.data.pageable.default-page-size` instead. 

I have removed `ApplicationConfiguration`.

sdelamo:
> * A Book entity is created but then not used anywhere in the example, nor is association handling demonstrated or tested anywhere. Why is it there?

I have removed the `Book` Entity.  It was there originally because we had such domain composition in a previous guide. And I think it is good to show some common one-to-many mapping. We can do that in a different guide.

sdelamo:
> * Why create command objects like GenreSaveCommand which only have a single property? Is there some functional reason for this?

I personally prefer to encapsulate the HTTP Requests' body in a POJO. Do you want me to use method parameters instead of POJOs?

guides/micronaut-data-jdbc-repository/micronaut-data-jdbc-repository.adoc
sdelamo: I've added the link.

graemerocher:
> I personally prefer to encapsulate the HTTP Requests' body in a POJO. Do you want me to use method parameters instead of POJOs?

This pattern makes sense if you have several properties, but in the case where there is only one it is just noise and more verbosity

sdelamo:
> > I personally prefer to encapsulate the HTTP Requests' body in a POJO. Do you want me to use method parameters instead of POJOs?
> 
> This pattern makes sense if you have several properties, but in the case where there is only one it is just noise and more verbosity

I've deleted `GenreCreateCommand` which had only one field. Now, the method signature looks like: 

`public HttpResponse<Genre> save(@Body("name") @NotBlank String name) {`

sdelamo:
@graemerocher can I merge this?

graemerocher (approved):

PR #275, opened 2021-06-07
Oracle Cloud Function guide

burtbeckwith:


burtbeckwith:
Weird test failure in the two Groovy projects, caused by NPE at line 80 of DefaultBeanContext:

`protected static final Logger LOG_LIFECYCLE = LoggerFactory.getLogger(DefaultBeanContext.class.getPackage().getName() + ".lifecycle");`

Any thoughts?

guides/micronaut-oracle-function/java/src/main/java/example/micronaut/Function.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Date;
+
+@Singleton
+public class Function extends OciFunction {
+
+    @Inject
+    TenancyIdProvider tenantIdProvider;
+
+    @ReflectiveAccess
+    public String handleRequest() {
```
ilopmar: Why do we need `@ReflectiveAccess` here?

ilopmar (changes requested):

guides/micronaut-oracle-function/java/src/main/java/example/micronaut/Function.java
graemerocher: Because Project.fn invokes the function handler via reflection, which is beyond our control

ilopmar: I though that the annotation was only necessary for GraalVM, but the guide doesn't use GraalVM.
In any case I think adding an explanation in the guide about why is necessary is good for users.

graemerocher: The guide needs to be updated to cover native image as well

ilopmar (approved):

PR #291, opened 2021-06-09
run only tests for changed guides

sdelamo:
This will allow us to run only the tests for the guides you have changed in the PR. It runs the tests only if your changes are actually a code change. 

I want whole test suite to be run for cron github actions hence the check for the Github workflow name.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,14 +108,15 @@ task generateTestScript {
         output.createNewFile()
         StringBuilder sout = new StringBuilder()
         StringBuilder serr = new StringBuilder()
-        def proc = "git diff --name-only HEAD HEAD~1".execute()
+        def proc = "git diff --name-only origin/master".execute()
         proc.consumeProcessOutput(sout, serr)
         proc.waitForOrKill(1000)
         String err = serr.toString()
         if (err) {
             logger.error(err)
         }
-        boolean onlyImagesOrMarkdownOrAsciidocChanged = sout.toString().split("\n").every {String str ->
+        String[] changedFiles = sout.toString().split("\n")
+        boolean onlyImagesOrMarkdownOrAsciidocChanged = changedFiles.every {String str ->
             str.endsWith('adoc') ||
                     str.endsWith('md')   ||
                     str.endsWith('png')  ||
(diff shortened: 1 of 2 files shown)


ilopmar:
I don't understand this change. Do you want to run only the tests in a cronjob instead for PRs and when we merge on master?

sdelamo:
If I write a new guide or change an existing guide, I want o run only the tests for that guide not the tests of every guide.

ilopmar:
and what about what you said here?

> I want whole test suite to be run for cron github actions hence the check for the Github workflow name.

sdelamo:
> and what about what you said here?
> 
> > I want whole test suite to be run for cron github actions hence the check for the Github workflow name.

I want to run every tests in the cron job. 

However, when I do a PR and add a new guide or change a guide I want to run only the tests for that guide.

sdelamo:
You can see an example here: 

https://github.com/micronaut-projects/micronaut-guides/runs/2784118085?check_suite_focus=true

ilopmar:
The problem I see with this approach is that we need to go and check if there has been an error or not. For example, the snapshot has been failing for a while for Micronaut 3.0.x https://github.com/micronaut-projects/micronaut-guides/actions/workflows/gradle-snapshot.yml but unless we go there and see what's going on, we don't really now.

I'm also worried about the PRs that upgrade Micronaut version. Will all the test be executed for those PRs?

I think it's safer (and slower) to run all the tests.

sdelamo:
> I'm also worried about the PRs that upgrade Micronaut version. Will all the test be executed for those PRs?

I have done a commit to ensure in PRs which modifies `version.txt` or `pom.xml` is changed every tests is executed.

sdelamo:
> The problem I see with this approach is that we need to go and check if there has been an error or not. For example, the snapshot has been failing for a while for Micronaut 3.0.x https://github.com/micronaut-projects/micronaut-guides/actions/workflows/gradle-snapshot.yml but unless we go there and see what's going on, we don't really now.

I think this PR does not affect this. We should to investigate failures of crons jobs.

ilopmar:
> > The problem I see with this approach is that we need to go and check if there has been an error or not. For example, the snapshot has been failing for a while for Micronaut 3.0.x [micronaut-projects/micronaut-guides/actions/workflows/gradle-snapshot.yml](https://github.com/micronaut-projects/micronaut-guides/actions/workflows/gradle-snapshot.yml) but unless we go there and see what's going on, we don't really now.
> 
> I think this PR does not affect this. We should to investigate failures of crons jobs.

No, this PR doesn't affect it. It's been failing for a few weeks now. 

Last run failed because of missing `@Nonnull` dependency (I'll do the change and replace everything to use Micronaut `@NotNull` and `@NonNull`).

In the run from [two weeks ago](https://github.com/micronaut-projects/micronaut-guides/runs/2703536569?check_suite_focus=true#step:7:1512), these guides failed: 
```
  micronaut-database-authentication-provider-gradle-groovy
  micronaut-oauth2-cognito-gradle-java
  micronaut-oauth2-cognito-maven-java
  micronaut-oauth2-oidc-google-gradle-java
  micronaut-oauth2-oidc-google-maven-java
  micronaut-oauth2-linkedin-gradle-java
  micronaut-oauth2-linkedin-maven-java
  micronaut-oauth2-github-gradle-java
  micronaut-oauth2-github-maven-java
  micronaut-oauth2-okta-gradle-java
  micronaut-oauth2-okta-maven-java
```

ilopmar (approved):

PR #298, opened 2021-06-10
Added the Maven equivalent to the adding-commit-info guide

burtbeckwith:


ilopmar (commented):
Thanks for this @burtbeckwith :100: 

I've tested it and the `git.properties` file is created when executing the goals `verify`, `package`,... but it's not if I just do `./mvnw mn:run`.

sdelamo:
> Thanks for this @burtbeckwith 💯
> 
> I've tested it and the `git.properties` file is created when executing the goals `verify`, `package`,... but it's not if I just do `./mvnw mn:run`.

I think I have seen similar issue for Gradle in the past. If you package the app and run the fat JAR is the git.properties included ?

ilopmar:
> I think I have seen similar issue for Gradle in the past. If you package the app and run the fat JAR is the git.properties included ?

Yes

image

After that I can even run `./mvnw mn:run` and it works because the file is already there

buildSrc/src/main/java/io/micronaut/guides/feature/GitProperties.java
```diff
@@ -0,0 +1,78 @@
...
+                    .id("com.gorylenko.gradle-git-properties")
+                    .lookupArtifactId("gradle-git-properties")
+                    .build());
+            return;
+        }
+
+        String version = resolver.resolve("git-commit-id-plugin")
+                .orElseThrow(() -> new RuntimeException("git-commit-id-plugin not found in pom.xml"))
+                .getVersion();
+        String pluginXml = loadPluginXml().replace("@VERSION@", version);
```
sdelamo: We typically do this simpler. Instead of loading the plugin template and replacing the version we use the `BuildProperties`. See:  

https://github.com/micronaut-projects/micronaut-starter/blob/2.5.x/starter-core/src/main/java/io/micronaut/starter/feature/function/azure/AbstractAzureFunction.java#L92-L98

buildSrc/src/main/resources/git-commit-id-plugin.xml
```diff
@@ -0,0 +1,17 @@
+<plugin>
```
sdelamo: See above comment to replace `@VERSION@` usage.

guides/adding-commit-info/adding-commit-info.adoc
```diff
@@ -56,9 +56,43 @@ plugins {
 }
 ----
 
+:exclude-for-build:
+
+:exclude-for-build:gradle
+
+== Maven Git Commit Id Plugin
+
+If a `git.properties` file is available on the classpath, the http://docs.micronaut.io/latest/api/io/micronaut/management/endpoint/info/source/GitInfoSource.html[GitInfoSource] will expose the values in that file under the `git` key. Generation of a `git.properties` file will need to be configured as part of your build; for example, you may choose to use the https://github.com/git-commit-id/git-commit-id-maven-plugin[Maven Git Commit Id Plugin] plugin. The plugin provides a goal named `git-commit-id:revision` responsible for the `git.properties` file generation. It is automatically invoked upon the execution of the `classes` goal. You can find the generated file in the directory `target/classes`.
```
sdelamo: Instead of repeating this paragraph: I will move it outside of the build specific section. 

```
If a `git.properties` file is available on the classpath, the http://docs.micronaut.io/latest/api/io/micronaut/management/endpoint/info/source/GitInfoSource.html[GitInfoSource] will expose the values in that file under the `git` key. Generation of a `git.properties` file will need to be configured as part of your build;
```

guides/adding-commit-info/adding-commit-info.adoc
```diff
@@ -56,9 +56,43 @@ plugins {
...
+
+Modify `pom.xml` file to add the plugin:
+
+[source,xml]
+.pom.xml
+----
```
sdelamo, line 95: Maybe we should write a macro to load a plugin template into the asciidoc to avoid having it in two places. similar to what we do with `resource`.  thoughts?

sdelamo (changes requested):

burtbeckwith:
I configured the plugin to run in the `compile` phase, so it generates the properties file if you run `compile` or another goal that depends on it. `mn:run` does invoke compile, but apparently not in a way that also triggers git.properties generation. I'll wait to resolve this before merging

sdelamo (approved):

PR #300, opened 2021-06-11
remove jcenter

sdelamo:


ilopmar:
A change in the build should run all the tests because we may have broken something

PR #301, opened 2021-06-11
run mayor tests in 3.0.x branch

sdelamo:


ilopmar:
This will work but I've seen you've already changed the `version.txt` file in branch `3.0.x`. That will fail because the workflow already changes the version https://github.com/micronaut-projects/micronaut-guides/blob/3.0.x/.github/workflows/gradle-snapshot.yml#L56-L57. You need to remove that step in the `3.0.x` branch

ilopmar (approved):

PR #310, opened 2021-06-15
Add a minimumJavaVersion metadata property

burtbeckwith:
The https://github.com/micronaut-projects/micronaut-guides/pull/275 PR builds are failing in JDK 8 because fnproject.io only supports JDK 11+. This PR adds a `minimumJavaVersion` property that will allow skipping JDK 8 tests

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/guides/Utils.groovy
+++ b/buildSrc/src/main/groovy/io/micronaut/guides/Utils.groovy
@@ -1,8 +1,10 @@
 package io.micronaut.guides
 
+import groovy.transform.CompileStatic
 import io.micronaut.starter.options.JdkVersion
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/groovy/io/micronaut/guides/Utils.groovy
```diff
@@ -0,0 +1,36 @@
+package io.micronaut.guides
+
+import io.micronaut.starter.options.JdkVersion
+import org.gradle.api.GradleException
+
+class Utils {
```
ilopmar, line 8: This class probably can be `@CompileStatic`

ilopmar (approved):
Looks great! :tada:

PR #317, opened 2021-06-16
Upgrade guides to Micronaut 2.5.7

ilopmar:


ilopmar:
Woooooooooow!!! Finally all the tests passed after 6-7 runs... Now I'm afraid of merging this because I don't want to break master... :grimacing:

PR #325, opened 2021-06-24
simplify features by moving common logic to a base class

burtbeckwith:


ilopmar (approved):
Looks good, thanks!

@sdelamo I think we should re-think the "no run all the tests thing" we did because for a PR like this one I would like to run all the tests, and none of the run.

sdelamo:
> Looks good, thanks!
> 
> @sdelamo I think we should re-think the "no run all the tests thing" we did because for a PR like this one I would like to run all the tests, and none of the run.

Anything changing `buildSrc` should execute every test ?

sdelamo (approved):

ilopmar:
> Anything changing `buildSrc` should execute every test ?

I think so because if we break something related to the dependencies (for example this PR changes that), we don't know.

PR #332, opened 2021-06-29
Oracle Cloud HTTP API Gateway Function guide

burtbeckwith:
CI will fail until https://github.com/micronaut-projects/micronaut-guides/pull/317 is merged because the tests need v2.5.7+


graemerocher:
2.5.7 has been released

burtbeckwith:
The guides currently use 2.5.5 but the CI tests for the [2.5.7 upgrade](https://github.com/micronaut-projects/micronaut-guides/pull/317) have been failing - Ivan mentioned that in today's meeting. Once that's stable and merged I'll update this PR.

graemerocher:
Gotcha

ilopmar:
CI passed yesterday for 2.5.7 upgrade and this morning I've restarted it and it worked, so I've merged the PR. I've also merged master branch into this one and CI is running now.

buildSrc/src/main/resources/pom.xml
```diff
@@ -134,5 +134,13 @@
             <artifactId>micronaut-oraclecloud-atp</artifactId>
             <version>1.3.5</version>
         </dependency>
+
+        <!-- Oracle Cloud SDK Core -->
+        <dependency>
+            <groupId>com.oracle.oci.sdk</groupId>
+            <artifactId>oci-java-sdk-core</artifactId>
+            <version>1.34.0</version>
```
graemerocher: Is this needed? It seems to be declared in the `micronaut-oraclecloud-bom` module which is dependency managed

guides/micronaut-oracle-function-http/micronaut-oracle-function-http.adoc
```diff
@@ -0,0 +1,370 @@
...
+<4> This endpoint accepts `POST` requests and starts the specified instance if it is stopped
+<5> This endpoint accepts `POST` requests and stops the specified instance if it is running
+
+== Testing the Application
+
+NOTE: The code in this guide should work with recent versions of Micronaut 2.5+, but the tests require at least version 2.5.7
+
+We need to update the generated `MicronautguideControllerTest` to test the changes we made in `MicronautguideController`.
+
+This will be a unit test, so we'll need some mock beans to replace the beans auto-registered by the `oracle-cloud-sdk` module which make requests to Oracle Cloud. We also need a workaround for the strict classloader isolation used by https://fnproject.io/[Fn Project] (which powers Oracle Cloud Functions).
```
graemerocher: Maybe don't use the word `workaround` here, it seems that is the way `Fn` is designed. Rephrase to `We also need to take into consideration the strict class loader isolation...`

guides/micronaut-oracle-function-http/micronaut-oracle-function-http.adoc
```diff
@@ -0,0 +1,370 @@
...
+Create the `MockData` class:
+
+test:mock/MockData[]
+
+Next create the `MockAuthenticationDetailsProvider` class. The methods all return `null` since they won't be called; the bean merely needs to exist for dependency injection:
+
+test:mock/MockAuthenticationDetailsProvider[]
+
+<1> The `AuthCachingPolicy` annotation is used to disable caching; without this we would need to provide a valid private key since the provider methods would be invoked when constructing SDK client classes
+<2> We use `@Replaces` to replace a previously auto-registered `ConfigFileAuthenticationDetailsProvider` bean (which happens if $HOME/.oci/config exists)
```
graemerocher: `ConfigFileAuthenticationDetailsProvider` is just one possible auth method, you could be using instance principals or resource principals so just rephrase to `... to replace the previously auto-configuration Oracle Cloud authentication method`

graemerocher (changes requested):

ilopmar:
JDK 11 build failed again because of some test failing randomly in the zipkin guide. I've pushed a draft PR to run the test with 2.5.6 (https://github.com/micronaut-projects/micronaut-guides/pull/333) so we can see in which version we've broken this.

buildSrc/src/main/resources/pom.xml
burtbeckwith: The pom.xml file isn't used as a pom file, it's parsed at https://github.com/micronaut-projects/micronaut-starter/blob/2.5.x/starter-core/src/main/java/io/micronaut/starter/build/dependencies/PomDependencyVersionResolver.java#L46 to lookup the version for a Feature, in this case https://github.com/micronaut-projects/micronaut-guides/blob/micronaut-oracle-function-http/buildSrc/src/main/java/io/micronaut/guides/feature/OracleCloudSdkCore.java. This keeps all the versions in one place. We could probably add a flag to indicate that the version is specified in a bom and not do the lookup though

burtbeckwith: Ignore the previous comment, you're right it isn't needed

graemerocher (approved):

PR #337, opened 2021-07-01
Added "Copy to Clipboard" button to code blocks

### LINKED ISSUE #245: Add "Copy to clipboard" option like in Micronaut docs
Add a "Copy to clipboard" button in the guides like the one existing in the Micronaut documetation that copies the code snippet to the clipboard:

image

burtbeckwith:
Fixes #245

### DIFF
--- a/assets/js/guide.js
+++ b/assets/js/guide.js
@@ -1,0 +1,43 @@
+function createCopyToClipboardElement() {
+    var copyToClipboardDiv = document.createElement('div');
+    var copyToClipboardSpan = document.createElement('span');
+    copyToClipboardSpan.setAttribute('class', 'copytoclipboard');
+    copyToClipboardSpan.setAttribute('onclick', 'copyToClipboard(this);');
+    copyToClipboardSpan.innerText = 'Copy to Clipboard';
+    copyToClipboardDiv.appendChild(copyToClipboardSpan);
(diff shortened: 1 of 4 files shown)


ilopmar (changes requested):
This changes don't work for me, both the index page and any individual guide page are "black":

image


There is no javascript error in the console, thought.

I've tried with Brave & Firefox.

burtbeckwith:
@ilopmar could you try on another machine? I did the work on Linux and tested in Firefox, but it works fine in Chrome and also in Chrome and Firefox on Mac. The code is from the user guide (multi-language-sample.js and guide.js) - the only change was the element where the button is added, `listingblock` instead of `multi-language-sample` since we only show one language per guide

burtbeckwith:
@ilopmar I see the problem - the page runs differently when you load it with a `file://` url than when it's served from a web server. I use livereload (`cd build/dist && livereload -p 3000`) but any server will work

burtbeckwith:
It should work now, I was using `<script />` but it needs to be `<script></script>`

ilopmar:
Yes, now it works. I've tried serving the content with python (`sudo python3 -m http.server 80 --directory build/dist`) and it didn't work either. Now it also works with you last commit even if I use `file://....`

ilopmar (approved):

PR #338, opened 2021-07-02
Cleanup

burtbeckwith:
deleted unused images
deleted unused snippets
added @api@ replacement with value 'https://docs.micronaut.io/latest/api to simplify javadoc links
phrasing, minor cleanup/fixes


ilopmar (approved):
Thanks Burt. I've reverted the deletion of some images that were still used (in the index and in one guide)

ilopmar (approved):

PR #359, opened 2021-07-13
Add Groovy and Kotlin code to RabbitMQ guide

burtbeckwith:


buildSrc/src/main/groovy/io/micronaut/guides/GuideProjectGenerator.groovy
```diff
@@ -126,12 +126,17 @@ class GuideProjectGenerator implements Closeable {
             }
 
             for (App app: metadata.apps) {
-                List<String> appFeatures = app.features
+                List<String> appFeatures = [] + app.features
```
sdelamo, line 129: why is this necessary?

guides/micronaut-rabbitmq/analytics/groovy/src/main/groovy/example/micronaut/AnalyticsController.groovy
```diff
@@ -0,0 +1,21 @@
...
+@Controller('/analytics')
+class AnalyticsController {
+
+    private final AnalyticsService analyticsService
+
+    AnalyticsController(AnalyticsService analyticsService) {
+        this.analyticsService = analyticsService
+    }
+
+    @Get('/') // <1>
```
sdelamo, line 17: This could be `@Get // <1>`

guides/micronaut-rabbitmq/analytics/groovy/src/main/groovy/example/micronaut/AnalyticsController.groovy
```diff
@@ -0,0 +1,21 @@
...
+    List<BookAnalytics> listAnalytics() {
+        return analyticsService.listAnalytics()
```
sdelamo, line 19: we could remove the optional `return`

ilopmar (approved):

guides/micronaut-rabbitmq/analytics/groovy/src/main/groovy/example/micronaut/AnalyticsService.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+@CompileStatic
+@Singleton
+class AnalyticsService {
+
+    private final Map<Book, Long> bookAnalytics = new ConcurrentHashMap<>() // <1>
+
+    void updateBookAnalytics(Book book) { // <2>
+        bookAnalytics.compute(book, (k, v) -> {
+            if (v == null) {
```
sdelamo, line 16: Maybe rewrite this: 

```groovy
            if (v == null) {
                 return 1L
             } else {
                 return v + 1
             }            
```


as: 

`v == null ? 1L : v + 1`

guides/micronaut-rabbitmq/analytics/groovy/src/main/groovy/example/micronaut/AnalyticsService.groovy
```diff
@@ -0,0 +1,27 @@
...
+                return 1L
+            } else {
+                return v + 1
+            }
+        })
+    }
+
+    List<BookAnalytics> listAnalytics() { // <3>
+        return bookAnalytics.collect { e -> new BookAnalytics(e.key.isbn, e.value) }
```
sdelamo, line 25: Maybe, remove `return` keyword.

guides/micronaut-rabbitmq/analytics/groovy/src/main/resources/application.yml
```diff
@@ -0,0 +1,18 @@
+micronaut:
+  application:
+    name: analytics
+---
+#tag::port[]
+micronaut:
+  server:
+    port: 8081 # <1>
```
sdelamo, line 8: I think this should use such a strategy: https://github.com/micronaut-projects/micronaut-guides/pull/353

guides/micronaut-rabbitmq/analytics/groovy/src/test/groovy/example/micronaut/AnalyticsServiceSpec.groovy
```diff
@@ -0,0 +1,40 @@
...
+        3 == findBookAnalytics(b1, analytics).count
+        1 == findBookAnalytics(b2, analytics).count
+    }
+
+    private BookAnalytics findBookAnalytics(Book b, List<BookAnalytics> analytics) {
+        BookAnalytics bookAnalytics = analytics.find { it.bookIsbn == b.isbn }
+        if (!bookAnalytics) {
+            throw new RuntimeException('Book not found')
+        }
+        return bookAnalytics
```
sdelamo, line 38: remove `return`

guides/micronaut-rabbitmq/analytics/kotlin/src/main/resources/application.yml
```diff
@@ -0,0 +1,18 @@
+micronaut:
+  application:
+    name: analytics
+---
+#tag::port[]
+micronaut:
+  server:
+    port: 8081 # <1>
```
sdelamo, line 8: I think we should do something like: https://github.com/micronaut-projects/micronaut-guides/pull/353

sdelamo (changes requested):
I know @ilopmar merged this PR but I think we should do at least the port changes using default environment. see: https://github.com/micronaut-projects/micronaut-guides/pull/353

buildSrc/src/main/groovy/io/micronaut/guides/GuideProjectGenerator.groovy
burtbeckwith: I saw this when removing the mockito feature for Spock - if Groovy is processed before Kotlin or Java, the removal affects those too. This wasn't noticed for GraalVM because the build still compiles and tests pass, but it was causing the Java/Kotlin apps to fail because they didn't have the mockito dependency. Making a copy for each loop iteration makes the change local to that loop iteration and doesn't affect Java or Kotlin, only the current Groovy app

PR #396, opened 2021-07-28
Show diff link for users with existing apps

### LINKED ISSUE #375: Consider Linking To Diff In Launch
Guides generally include the command that would create a starting point application with features related to the app being built in the guide.  That might be something like this:

    mn create-app --features asciidoctor,camunda,openapi demo

Consider adding a link to the guide which shows the diff between a default created app and an app created with the specified features, similar to the Diff capability that is available in Launch.

Launch supports a link like this which pre-populates the form with the specified features and config options:

https://micronaut.io/launch?javaVersion=JDK_11&lang=JAVA&build=GRADLE&test=JUNIT&name=demo&package=com.example&type=DEFAULT&features=asciidoctor&features=openapi&features=camunda&version=2.5.9

From there the user can click the Diff button to see what was added to the project as a result of specifying the requested features and config options.

It might be handy to provide a link from the guide straight to the diff, like this:

https://micronaut.io/launch?javaVersion=JDK_11&lang=JAVA&build=GRADLE&test=JUNIT&name=demo&package=com.example&type=DEFAULT&features=asciidoctor&features=openapi&features=camunda&version=2.5.9&activity=diff

Alternatively, maybe the diff could be generated at Guide build time and included in the generated artifacts as a kind of JS popup or similar.

Thanks for your consideration!

burtbeckwith:
Adds a `diffLink` macro that will open a diff link of the form `https://micronaut.io/launch?features=...&lang=...&build=...&test=...&name=...&type=...&package=example.micronaut&activity=diff` so users can apply the changes resulting from the selected features to an existing application.

difflink

Fixes https://github.com/micronaut-projects/micronaut-guides/issues/375

### DIFF
--- a/guides/micronaut-database-authentication-provider/micronaut-database-authentication-provider.adoc
+++ b/guides/micronaut-database-authentication-provider/micronaut-database-authentication-provider.adoc
@@ -23,9 +23,7 @@ If you use Micronaut Launch, select Micronaut Application as application type an
 
 include::{commondir}/common-default-package.adoc[]
 
-NOTE: If you have an existing Micronaut application and want to add the functionality described here, you can
 diffLink:[featureExcludes=spring-security-crypto]
-and apply those changes to your application.
 
 include::{commondir}/common-annotationprocessors.adoc[]
 
--- a/guides/micronaut-database-authentication-provider/micronaut-database-authentication-provider.adoc
+++ b/guides/micronaut-database-authentication-provider/micronaut-database-authentication-provider.adoc
@@ -23,6 +23,10 @@ If you use Micronaut Launch, select Micronaut Application as application type an
 
 include::{commondir}/common-default-package.adoc[]
 
+NOTE: If you have an existing Micronaut application and want to add the functionality described here, you can
+diffLink:[featureExcludes=spring-security-crypto]
+and apply those changes to your application.
+
 include::{commondir}/common-annotationprocessors.adoc[]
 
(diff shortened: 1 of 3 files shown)


guides/micronaut-database-authentication-provider/micronaut-database-authentication-provider.adoc
```diff
@@ -23,6 +23,10 @@ If you use Micronaut Launch, select Micronaut Application as application type an
 
 include::{commondir}/common-default-package.adoc[]
 
+NOTE: If you have an existing Micronaut application and want to add the functionality described here, you can
+diffLink:[featureExcludes=spring-security-crypto]
+and apply those changes to your application.
+
```
ilopmar: Looks good. The only improvement I'd add would be moving everything to the macro so we can add it easily to all the guides doing just `diffLink:[featureExcludes=spring-security-crypto]` and that generates the whole copy:

```
NOTE: If you have an existing Micronaut application and want to add the functionality described here, you can
https://micronaut.io/launch?features=security-jwt&features=security-ldap&features=hibernate-gorm&lang=GROOVY&build=GRADLE&test=SPOCK&name=micronautguide&type=DEFAULT&package=example.micronaut&activity=diff[view the dependency and configuration changes from the specified features, window="_blank"]
and apply those changes to your application.
```

ilopmar (approved):
Now looks perfect!

PR #401, opened 2021-07-30
java: use public modifiers

sdelamo:
In other guides, we use `public` modifier for Java controller classes and their methods annotated with @GET... 

Do we want to change the style to use package private scope ?


burtbeckwith:
I'm fine either way but I prefer to omit `public` in controllers since it's not needed. Since they're only invoked by the generated code that's in the same package, omitting them in the controllers is a small reminder that controllers aren't injectable or directly invocable from user code.

graemerocher:
I agree with Burt

sdelamo:
Please do a vote here: https://github.com/micronaut-projects/micronaut-guides/discussions/403

PR #408, opened 2021-07-30
add @NonNull annotations to fields

sdelamo:
I typically write the class with all the fields annotated with `@NonNull` annotations (`io.micronaut.core.annotations`) and also the constraint annotations (`@NotBlank`, `@Size`) and then use IntelliJ to generate getters and setters and constructors which automatically adds the nullability annotations to the getters and setters. Do you think we should not have micronaut nullability annotations in fields?

I have also changed from `@NotNull` to `@NotBlank` for Strings


burtbeckwith (approved):
Looks good, thanks

PR #410, opened 2021-08-01
Use brand guidelines 

### LINKED ISSUE #398: Updating guides
As we are working through making sure all of our publications are written in line with our brand guidelines, we noticed there are several uses of the word "Micronaut" as a noun in our guides.
“Micronaut” should be used as an adjective and we have some guidelines for how we need to write it in sentences.
Here is a link to the brand guidelines that should help https://micronaut.io/brand-guidelines/.

Please also see some examples below:

Incorrect - Micronaut supports the concept of one or many default environments. A default environment is one that is only applied if no other environments are explicitly specified or deduced.
Correct - The Micronaut framework supports the concept of one or many default environments. A default environment is one that is only applied if no other environments are explicitly specified or deduced.

Incorrect - Read more about this support with Micronaut.
Correct - Read more about this support with the Micronaut framework.

Incorrect - Micronaut will automatically convert it to JSON before sending it.
Correct - The Framework will automatically convert it to JSON before sending it.
If you feel like you are saying the Micronaut framework a lot, you can use Framework with an uppercase “F” when referring to the Micronaut framework.

We really appreciate your help and understanding as we are navigating through this.
Thanks team!

burtbeckwith:
Fixes https://github.com/micronaut-projects/micronaut-guides/issues/398


sdelamo (approved):
Thanks for doing this change.

PR #422, opened 2021-08-07
feat: logstash guide

sdelamo:
This is draft PR to show how to use distributed logging with [Logstash](https://www.elastic.co/logstash/) and a Micronaut Application. 

To generate the guide locally: 

`./gradlew -Dmicronaut.guide=micronaut-logstash build`

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/guides/IndexGenerator.groovy
+++ b/buildSrc/src/main/groovy/io/micronaut/guides/IndexGenerator.groovy
@@ -275,7 +275,7 @@ class IndexGenerator {
     @CompileDynamic
     private static List<String> generateTags(GuideMetadata guide) {
         [
-            guide.tags +
+            guide.tags ?: [] +
             guide.languages*.toString() +
             guide.buildTools*.toString() +
             guide.apps.collect { it.features }.flatten().unique()
--- a/guides/micronaut-logstash/micronaut-logstash.adoc
+++ b/guides/micronaut-logstash/micronaut-logstash.adoc
@@ -1,0 +1,58 @@
+include::{commondir}/common-header.adoc[]
+
+
+
+include::{commondir}/common-requirements.adoc[]
+
+include::{commondir}/common-completesolution.adoc[]
+
+
+== Install Logstash
+
+https://www.elastic.co/logstash/[Logstash]:
+
+____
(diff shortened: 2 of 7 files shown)


buildSrc/src/main/groovy/io/micronaut/guides/IndexGenerator.groovy
```diff
@@ -275,7 +275,7 @@ class IndexGenerator {
     @CompileDynamic
     private static List<String> generateTags(GuideMetadata guide) {
         [
-            guide.tags +
+            guide.tags ?: [] +
```
ilopmar, line 278: I think it's better to remove this so we make sure all the guides have at least one tag (besides the ones added automatically)

guides/micronaut-logstash/micronaut-logstash.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+https://www.elastic.co/logstash/[Logstash]:
+
+____
+Logstash is a free and open server-side data processing pipeline that ingests data from a multitude of sources, transforms it, and then sends it to your favorite "stash."
+____
+
+* https://www.elastic.co/downloads/logstash[Download and unzip Logstash]
+* Prepare a https://www.elastic.co/guide/en/logstash/current/configuration.html[`logstash.conf` config file]
+* Run `bin/logstash -f logstash.conf`
```
ilopmar, line 20: Even if we want to keep this I think we should provide the `docker run ...` command to execute Logstash in a Docker container.

ilopmar (changes requested):

buildSrc/src/main/groovy/io/micronaut/guides/IndexGenerator.groovy
sdelamo: I don't think we should force a guide to have a tag, I don't see any harm on making tags optional.

guides/micronaut-logstash/micronaut-logstash.adoc
sdelamo: Note, this is a draft. It is not a finished guide. I am just dumping some experiments in a draft PR.

ilopmar: :+1:

Issue #444, opened 2021-08-19
Generate Twitter Cards automatically


sdelamo:
[Twitter Cards](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/abouts-cards)

PR #446, opened 2021-08-19
Applescript to generate twitter card

sdelamo:
see https://github.com/micronaut-projects/micronaut-guides/issues/444

generate-twitter-card-for-micronaut


ilopmar (approved):
I can't really test this because I'm on Linux, but I'll assume it works :-)

PR #448, opened 2021-08-19
feat: cli application to generate a list of tweets from the guides

### LINKED ISSUE #447: CLI to generate tweets about guides

sdelamo:


cli/src/main/java/io/micronaut/guides/TweetsCommand.java
```diff
@@ -0,0 +1,62 @@
...
+import jakarta.inject.Inject;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+import reactor.core.publisher.Mono;
+import java.util.ArrayList;
+import java.util.List;
+
+@Command(name = "guidetweets", description = "generates a list of tweets", mixinStandardHelpOptions = true)
+public class TweetsCommand implements Runnable {
+    private static final String HASHTAG = " #";
```
ilopmar: Why do you need the extra space before the `#`. There is already one space in the generated text:

```
📖 Micronaut Guide: Micronaut JWT Authentication https://guides.micronaut.io/latest/micronaut-security-jwt.html  #micronaut
```

ilopmar (commented):
Beside my comment, why don't you use the starter BuildTool, Language...? Just for not adding the dependency and making the final artifact bigger?

ilopmar:
One other thing I missed. Please add a simple `README.md` that explains how to build and how to run the tool. Something like:

```
Build:
./gradlew ....

Run
java -jar .....

Run without build
./gradlew cli:run --args="--hashtag=myHashtag --prefix=foo"
```

ilopmar (approved):

PR #456, opened 2021-08-20
micronaut-security-session guide fixes

burtbeckwith:
I was working through the guide but the code wasn't compiling because `reactor` wasn't in the guide as a feature/dependency. Did some more fixes/cleanup while I was there.


guides/micronaut-security-session/groovy/src/main/groovy/example/micronaut/HomeController.groovy
```diff
@@ -13,6 +15,7 @@ import java.security.Principal
 @Controller("/") // <2>
 class HomeController {
 
+    @Produces(MediaType.TEXT_HTML)
```
sdelamo: It should not be necessary. When you specify `@View` HTML content type is inferred.

guides/micronaut-security-session/groovy/src/main/groovy/example/micronaut/LoginAuthController.groovy
```diff
@@ -12,15 +14,17 @@ import io.micronaut.views.View
 @Controller("/login") // <2>
 class LoginAuthController {
 
+    @Produces(MediaType.TEXT_HTML)
     @Get("/auth") // <3>
     @View("auth") // <4>
     Map<String, Object> auth() {
-        [:]
+        Collections.emptyMap()
```
sdelamo: why not use `[:]' ? To avoid instantiating a new empty Hash Map ?

guides/micronaut-security-session/micronaut-security-session.adoc
```diff
@@ -2,7 +2,7 @@ include::{commondir}/common-header-top.adoc[]
 
 == Getting Started
 
-In this guide you will write a Micronaut application with session based authentication.
+In this guide, we will create a Micronaut application written in @language@ with session based authentication.
```
sdelamo, line 5: I think I it is better to direct the guides to the reader. `you` vs `we`

guides/micronaut-security-session/kotlin/src/test/groovy/example/micronaut/SessionAuthenticationSpec.groovy
```diff
@@ -1,79 +0,0 @@
...
-
-import geb.spock.GebSpec
-import io.micronaut.runtime.server.EmbeddedServer
-import io.micronaut.test.extensions.spock.annotation.MicronautTest
-
-import jakarta.inject.Inject
-
-//tag::clazz[]
-@MicronautTest // <1>
-class SessionAuthenticationSpec extends GebSpec {
```
sdelamo, line 11: why are we deleting the test?

guides/micronaut-security-session/java/src/test/groovy/example/micronaut/SessionAuthenticationSpec.groovy
```diff
@@ -1,79 +0,0 @@
...
-
-import geb.spock.GebSpec
-import io.micronaut.runtime.server.EmbeddedServer
-import io.micronaut.test.extensions.spock.annotation.MicronautTest
-
-import jakarta.inject.Inject
-
-//tag::clazz[]
-@MicronautTest // <1>
-class SessionAuthenticationSpec extends GebSpec {
```
sdelamo, line 11: Why are we deleting the test ?

sdelamo (changes requested):

guides/micronaut-security-session/groovy/src/main/groovy/example/micronaut/LoginAuthController.groovy
burtbeckwith: this is microoptimization, but `[:]` creates a `java.util.HashMap` with the default 10 slots (just like `new HashMap<>()`), and `Collections.emptyMap()` returns the singleton instance of `java.util.Collections.EmptyMap` which has no state.

guides/micronaut-security-session/groovy/src/main/groovy/example/micronaut/HomeController.groovy
burtbeckwith: I added that to be consistent with the Java version - should I remove @Produces from there?

guides/micronaut-security-session/micronaut-security-session.adoc
burtbeckwith: I copied that from common-gettingStarted.adoc - should I change we to you there?

guides/micronaut-security-session/kotlin/src/test/groovy/example/micronaut/SessionAuthenticationSpec.groovy
burtbeckwith: see https://github.com/micronaut-projects/micronaut-guides/pull/456/commits/5201b6470974aebbc750afa71e580014ea8b1a79

guides/micronaut-security-session/groovy/src/main/groovy/example/micronaut/HomeController.groovy
sdelamo: I think we should remove it everywhere we use `@Views` unless we want to return something different than HTML. 

`@Views` defaults to HTML content type.  see: https://github.com/micronaut-projects/micronaut-views/blob/master/views-core/src/main/java/io/micronaut/views/ViewsFilter.java#L124

If we have it in the java version, I think we should remove it from there as well.

guides/micronaut-security-session/micronaut-security-session.adoc
sdelamo: Let's discuss it separately.

guides/micronaut-security-session/java/src/main/java/example/micronaut/LoginAuthController.java
```diff
@@ -1,29 +1,24 @@
...
 @Secured(SecurityRule.IS_ANONYMOUS) // <1>
 @Controller("/login")  // <2>
 public class LoginAuthController {
 
-    @Produces(MediaType.TEXT_HTML)
     @Get("/auth") // <3>
     @View("auth") // <4>
     public Map<String, Object> auth() {
-        return new HashMap<>();
+        return Collections.emptyMap();
```
sdelamo: We should not return an inmutable Map as the model. Because, we support a feature called [ViewModelProcessors](https://micronaut-projects.github.io/micronaut-views/latest/guide/#custom) which allows you to decorate a model. That it is to say add stuff to the map.

guides/micronaut-security-session/kotlin/src/main/kotlin/example/micronaut/LoginAuthController.kt
```diff
@@ -12,13 +12,9 @@ class LoginAuthController {
 
     @Get("/auth") // <3>
     @View("auth") // <4>
-    fun auth(): Map<String, Any> {
-        return HashMap()
-    }
+    fun auth(): Map<String, Any> = emptyMap()
```
sdelamo: if Kotlin `emptyMap() `is immutable we should not do that. We should not return an inmutable Map as the model. Because, we support a feature called ViewModelProcessors which allows you to decorate a model. That it is to say add stuff to the map.

guides/micronaut-security-session/kotlin/src/main/kotlin/example/micronaut/LoginAuthController.kt
```diff
@@ -12,13 +12,9 @@ class LoginAuthController {
...
 
     @Get("/authFailed") // <5>
     @View("auth") // <4>
-    fun authFailed(): Map<String, Any> {
-        return mapOf(Pair("errors", true))
-    }
+    fun authFailed(): Map<String, Any> = mapOf("errors" to true)
```
sdelamo: not sure if this should be `mutableMapOf`. see reasons above.

sdelamo (changes requested):

PR #466, opened 2021-08-23
add reactor feature to micronaut-rabbitmq guide

burtbeckwith:
also remove unused reactor from analytics app

### DIFF
--- a/guides/micronaut-rabbitmq/metadata.json
+++ b/guides/micronaut-rabbitmq/metadata.json
@@ -14,7 +14,7 @@
     },
     {
       "name": "analytics",
-      "features": ["graalvm", "rabbitmq", "reactor"]
+      "features": ["graalvm", "rabbitmq"]
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


guides/micronaut-rabbitmq/metadata.json
```diff
@@ -14,7 +14,7 @@
     },
     {
       "name": "analytics",
-      "features": ["graalvm", "rabbitmq", "reactor"]
+      "features": ["graalvm", "rabbitmq"]
```
sdelamo, line 17: Do we need reactor for this guide? You add it in the copy but you remove it in the features list.

sdelamo (changes requested):

guides/micronaut-rabbitmq/metadata.json
burtbeckwith: It's only needed for the books microservice but not for analytics

PR #473, opened 2021-08-24
add reactor feature to micronaut-oauth2-github guide

burtbeckwith:


guides/micronaut-oauth2-github/java/src/main/resources/application.yml
```diff
@@ -21,7 +21,7 @@ micronaut:
             url: 'https://github.com/login/oauth/authorize' # <8>
           token:
             url: 'https://github.com/login/oauth/access_token' # <9>
-            auth-method: client-secret-post
+            auth-method: client_secret_post
```
sdelamo: Should we write CLIENT_SECRET_POST  for enums ? Not sure what is the best practice we want to recommend for such an enum https://micronaut-projects.github.io/micronaut-security/latest/api/io/micronaut/security/oauth2/endpoint/AuthenticationMethod.html

burtbeckwith: I don't have much of a preference - I changed this because IntelliJ was complaining that `client-secret-post` wasn't valid. But this is probably a bug since all 3 forms work in the running app.

sdelamo: If you are ok it, let's change to the uppercase version which matches the enum name and we err on the safe side.

sdelamo (approved):

PR #477, opened 2021-08-24
Kotlin and Groovy versions of code for micronaut-oauth2-linkedin guide

burtbeckwith:


guides/micronaut-oauth2-linkedin/groovy/src/main/groovy/example/micronaut/LinkedInMe.groovy
```diff
@@ -0,0 +1,34 @@
...
+
+    @NonNull
+    @NotBlank
+    final String localizedFirstName
+
+    @NonNull
+    @NotBlank
+    final String localizedLastName
+
+    @Creator
```
sdelamo: Is `@Creator` necessary? If you want to keep, we need to add a callout and explain it to the user.

guides/micronaut-oauth2-linkedin/java/src/main/java/example/micronaut/LinkedInMe.java
```diff
@@ -10,16 +11,17 @@
...
     @NotBlank
-    private String localizedFirstName;
+    private final String localizedFirstName;
 
     @NonNull
     @NotBlank
-    private String localizedLastName;
+    private final String localizedLastName;
 
+    @Creator
```
sdelamo: Is `@Creator` necessary? If you want to keep, we need to add a callout and explain it to the user.

guides/micronaut-oauth2-linkedin/kotlin/src/main/kotlin/example/micronaut/LinkedInMe.kt
```diff
@@ -0,0 +1,12 @@
+package example.micronaut
+
+import io.micronaut.core.annotation.Creator
+import io.micronaut.core.annotation.Introspected
+import io.micronaut.core.annotation.NonNull
+import javax.validation.constraints.NotBlank
+
+@Introspected
+data class LinkedInMe @Creator constructor(
```
sdelamo: Is `@Creator` necessary? If you want to keep, we need to add a callout and explain it to the user.

sdelamo (changes requested):

sdelamo (approved):

PR #478, opened 2021-08-24
add reactor feature to micronaut-security-jwt guide

burtbeckwith:


guides/micronaut-security-jwt/groovy/src/test/groovy/example/micronaut/AppClient.groovy
```diff
@@ -10,13 +10,16 @@ import io.micronaut.http.client.annotation.Client
...
 interface AppClient {
 
     @Post("/login")
     BearerAccessRefreshToken login(@Body UsernamePasswordCredentials credentials)
 
-    @Consumes(MediaType.TEXT_PLAIN)
+    @Consumes(TEXT_PLAIN)
     @Get
-    String home(@Header String authorization)
+    String home(@Header('authorization') String authorization)
```
sdelamo: Is this necessary ? If you want to be clear about the header name which will be used, we should use the constant in `HttpHeaders`.

guides/micronaut-security-jwt/kotlin/src/main/kotlin/example/micronaut/HomeController.kt
```diff
@@ -1,19 +1,18 @@
...
 class HomeController {
 
-    @Produces(MediaType.TEXT_PLAIN) // <3>
-    @Get("/")  // <4>
-    fun index(principal: Principal): String {  // <5>
-        return principal.name
-    }
+    @Produces(TEXT_PLAIN)
+    @Get // <3>
+    fun index(principal: Principal): String = principal.name // <4>
```
sdelamo, line 17: we got rid of the <5> callout. Did we remove it from the asciidoc as well?

sdelamo (changes requested):
It looks good. the changes to the Kotlin code are great. I have a couple of comments.

guides/micronaut-security-jwt/groovy/src/test/groovy/example/micronaut/AppClient.groovy
burtbeckwith: this was to be consistent with the Java and Kotlin code

guides/micronaut-security-jwt/kotlin/src/main/kotlin/example/micronaut/HomeController.kt
burtbeckwith: The Kotlin code had the extra callout, the Java and Groovy have 4 and the asciidoc has 4

guides/micronaut-security-jwt/groovy/src/test/groovy/example/micronaut/AppClient.groovy
sdelamo: Could we use 

`@Header String authorization`

or `@Header(HttpHeaders.AUTHORIZATION) String authorization` ?

I prefer the latter.

PR #521, opened 2021-09-14
use verb set up

sdelamo:
(setup is a noun; set up is the verb form)


sdelamo:
@kroegerk can you check this?

kroegerk (approved):
@sdelamo  There was one instance that did not need to be updated; I think I fixed it. The rest are correct.

PR #531, opened 2021-09-16
mongodb reactive

sdelamo:


guides/micronaut-mongodb-asynchronous/java/src/main/java/example/micronaut/MongoDbFruitRepository.java
```diff
@@ -0,0 +1,44 @@
...
+        this.mongoConf = mongoConf;
+        this.mongoClient = mongoClient;
+    }
+
+    @Override
+    public Mono<Boolean> save(@NonNull @NotNull @Valid Fruit fruit){
+        return Mono.from(getCollection().insertOne(fruit)) // <4>
+                .map(insertOneResult -> Boolean.TRUE)
+                .onErrorReturn(Boolean.FALSE);
+    }
```
ilopmar, line 31: Same comment as in the Mongo Sync guide. I think it is a common use case to have an `id` field assigned automatically Mongo and then return the `Fruit` object with that `id` field updated.

guides/micronaut-mongodb-asynchronous/java/src/test/java/example/micronaut/FruitControllerTest.java
```diff
@@ -0,0 +1,44 @@
...
+import io.micronaut.runtime.server.EmbeddedServer;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import java.util.Collections;
+import java.util.List;
+
+public class FruitControllerTest {
```
graemerocher, line 14: Does it make sense to share this template with the synchronous guide since they should be the same

ilopmar (commented):
One additional comment. Wouldn't it be better if you merge both guides into a single one and explain how to use both blocking and reactive clients?

graemerocher:
I personally kind of like that they are separate to explain best practise for each

guides/micronaut-mongodb-asynchronous/java/src/test/java/example/micronaut/FruitControllerTest.java
sdelamo: We don't have support for sharing code between guides.

sdelamo:
I prefer them to be separate. Shorter guides for the win.

sdelamo:
@kroegerk can you proofread it

kroegerk (approved):

PR #533, opened 2021-09-17
Proofreading

kroegerk:
A couple of minor adjustments for consistency

### DIFF
--- a/guides/micronaut-oauth2-client-credentials-cognito/micronaut-oauth2-client-credentials-cognito.adoc
+++ b/guides/micronaut-oauth2-client-credentials-cognito/micronaut-oauth2-client-credentials-cognito.adoc
@@ -113,9 +113,9 @@ Add the following OAuth2 configuration:
 resource:application.yml[app=bookrecommendation,tag=oauth2]
 
 
-<1> OAuth 2.0 Client name.
-<2> Client ID. See previous screenshot.
-<3> Client Secret. See previous screenshot.
+<1> OAuth 2.0 client name.
+<2> Client id. See previous screenshot.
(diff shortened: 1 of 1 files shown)


guides/micronaut-oauth2-client-credentials-cognito/micronaut-oauth2-client-credentials-cognito.adoc
```diff
@@ -113,9 +113,9 @@ Add the following OAuth2 configuration:
 resource:application.yml[app=bookrecommendation,tag=oauth2]
 
 
-<1> OAuth 2.0 Client name.
-<2> Client ID. See previous screenshot.
-<3> Client Secret. See previous screenshot.
+<1> OAuth 2.0 client name.
+<2> Client id. See previous screenshot.
```
burtbeckwith, line 117: I'd leave it as ID since it's an abbreviation

kroegerk: I'm fine with that. I had noticed there was a paragraph earlier in the article in which it was lowercase, and I was attempting to keep things consistent. Feel free to change all appearances to ID (in caps) if you feel that makes more sense. 

Same with things like "Client Name" and "Client Secret." The first use was in lowercase, so I adjusted the other instances accordingly. They could just as easily all be changed to title case. Just as long as they're consistent, I'm fine with either treatment.

sdelamo: Ok, so I will close this PR then.

PR #574, opened 2021-10-14
Don't generate projects which require 11 if running with 8

sdelamo:
Publish only when 11 which include every project.


ilopmar:
I think this is not the solution. I would like to know why the Oracle guides that only use JDK11 work and the Google Cloud Function doesn't

sdelamo:
> I think this is not the solution. I would like to know why the Oracle guides that only use JDK11 work and the Google Cloud Function doesn't

My guess is because starter does allow you to generate a JDK 8 project with Oracle. However, starter throws an exception if you do GCP and JDK 8

ilopmar (approved):

PR #595, opened 2021-10-21
skip eureka guide tests

sdelamo:
micronaut-projects/micronaut-discovery-client#305


ilopmar:
Can you please file an issue in this repo so we don't forget to enable the tests again?

PR #651, opened 2021-12-01
Update micronaut-security-jwt.adoc

psyklopz:
The setting `micronaut.security.token.jwt.generator.access-token-expiration` has been deprecated in favor of `micronaut.security.token.jwt.generator.access-token.expiration` since Micronaut 2.x.  (Notice the period in place of the dash.)

It seems that there may be a bug where the deprecated setting is no longer observed, but the new setting seems to work just fine.
I'm running Micronaut 2.5.11.

Regardless of the possible bug, it's probably best to keep the docs up-to-date and not recommend using deprecated settings.


sdelamo (approved):
@psyklopz  Thanks for the fix!

PR #660, opened 2021-12-08
build: update to Micronaut 3.2.1

sdelamo:


ilopmar:
The Gradle Git plugin only supports JDK11 now: https://github.com/micronaut-projects/micronaut-guides/runs/4458447118?check_suite_focus=true#step:7:18

```
An exception occurred applying plugin request [id: 'com.gorylenko.gradle-git-properties', version: '2.3.1']
> Failed to apply plugin 'com.gorylenko.gradle-git-properties'.
   > Could not create an instance of type com.gorylenko.GitPropertiesPluginExtension.
      > org/eclipse/jgit/api/errors/JGitInternalException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

@sdelamo I think we should use a previous version of the plugin for the guides that works with JDK 8, and file an issue here to upgrade the plugin when we upgrade Micronaut to JDK11.

sdelamo:
@ilopmar 

> The Gradle Git plugin only supports JDK11 now: https://github.com/micronaut-projects/micronaut-guides/runs/4458447118?check_suite_focus=true#step:7:18
> 
> ```
> An exception occurred applying plugin request [id: 'com.gorylenko.gradle-git-properties', version: '2.3.1']
> > Failed to apply plugin 'com.gorylenko.gradle-git-properties'.
>    > Could not create an instance of type com.gorylenko.GitPropertiesPluginExtension.
>       > org/eclipse/jgit/api/errors/JGitInternalException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
> ```
> 
> @sdelamo I think we should use a previous version of the plugin for the guides that works with JDK 8, and file an issue here to upgrade the plugin when we upgrade Micronaut to JDK11.

Should not we change the guide to be only JDK 11 and above instead?

ilopmar:
> Should not we change the guide to be only JDK 11 and above instead?

Yes, we can do that too. My point is that users can still use JDK 8 and the Gradle Git plugin if we use another version of the plugin. This is not the same when we use a library that only works with JDK11 (for example the GCP)

sdelamo:
I set the guide to have a minimum of JDK 11.

PR #667, opened 2021-12-16
A class needs to be `Serializable`.

### LINKED ISSUE #666: Cache Guide doesn't mention that class needs to be Serializable

jacekgajek:
Cache Guide didn't mention that class needs to be Serializable

Fix https://github.com/micronaut-projects/micronaut-guides/issues/666

### DIFF
--- a/guides/micronaut-cache/groovy/src/main/groovy/example/micronaut/News.groovy
+++ b/guides/micronaut-cache/groovy/src/main/groovy/example/micronaut/News.groovy
@@ -7,7 +7,7 @@ import java.time.Month
 
 @CompileStatic
 @Introspected // <1>
-class News {
+class News implements Serializable { // <2>
     Month month
     List<String> headlines
 }
--- a/guides/micronaut-cache/java/src/main/java/example/micronaut/News.java
+++ b/guides/micronaut-cache/java/src/main/java/example/micronaut/News.java
@@ -2,11 +2,12 @@ package example.micronaut;
 
 import io.micronaut.core.annotation.Introspected;
 
+import java.io.Serializable
 import java.time.Month;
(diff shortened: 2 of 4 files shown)


jacekgajek:
(I've signed the agreement it's just stuck on "pending")

sdelamo:
How can I verify this? I don't think it needed to be Serializable when we wrote the guide. Moreover, the guide has tests.

jacekgajek:
https://github.com/jacekgajek/redis-error-demo
This is actually a showcase for a bug in caching suspending funs, but if you remove `Serializable` then the non-suspending controllers will fail with error
```
 "message": "Internal Server Error: I/O error occurred during serialization: com.example.BarController$Bar"
```

jacekgajek:
Putting this here to prove that I agreed to the CLA
image

jacekgajek:
https://github.com/jacekgajek/redis-error-demo
This is actually a showcase for a bug in caching suspending funs, but if you remove `Serializable` then the non-suspending controllers will fail with error
```
 "message": "Internal Server Error: I/O error occurred during serialization: com.example.BarController$Bar"
```

sdelamo:
> https://github.com/jacekgajek/redis-error-demo This is actually a showcase for a bug in caching suspending funs, but if you remove `Serializable` then the non-suspending controllers will fail with error
> 
> ```
>  "message": "Internal Server Error: I/O error occurred during serialization: com.example.BarController$Bar"
> ```

Is this unrelated to the guide? If it is not related I would like to move it to a different issue. 

For the guide, we have tests. I don't think it is necessary. Am I missing anything?

sdelamo:
Closing this PR, I don't think it needs to be serializable.

PR #677, opened 2022-01-19
Add Kotlin source for micronaut-file-download-excel

burtbeckwith:


guides/micronaut-file-download-excel/java/src/main/java/example/micronaut/BookExcelServiceImpl.java
```diff
@@ -35,11 +34,12 @@ public SystemFile excelFileFromBooks(@NonNull @NotNull List<@Valid Book> bookLis
                                     cd.style(BookExcelStylesheet.STYLE_HEADER);
                                 })
                             ));
-                    bookList.stream()
-                            .forEach( book -> s.row(r -> {
-                                r.cell(book.getIsbn());
-                                r.cell(book.getName());
-                            }));
+                    for (Book book : bookList) {
```
sdelamo, line 37: Any reason you don't want to use the stream api?

sdelamo (approved):

guides/micronaut-file-download-excel/java/src/main/java/example/micronaut/BookExcelServiceImpl.java
burtbeckwith: why build a stream and a lambda when you can just loop through the list?

PR #684, opened 2022-01-25
Support partial guides

burtbeckwith:
rework to support composed guides, e.g. a "core" guide for deploying a cloud app with multiple cloud-specific variants that include the core guide and add deployment info for that cloud

these partial guide asciidoc files will be in `src/docs/common` so I moved existing common snippet files and callout files to their own subdirectories (`src/docs/common/snippets` and `src/docs/common/callouts`) and the partial files will be in `src/docs/common/partial`

also simplified the snippet include syntax from `include::{commondir}/common-header-top.adoc[]` to `common:header-top.adoc[]`

best to look at individual commits


ilopmar (commented):
It looks good. I've tested locally and opened a few guides and they look ok. 

I'm a little bit worried that we may over-complicate how to write a guide with different "fragments" (snippets, callouts, partials,...) coming from different places that will make difficult to figure out the guide structure from the asciidoctor file.

I'll leave the final review to Sergio.

sdelamo:
there are a lot of changes here I need some time. I will probably split this PR in multiple PRs.

ilopmar:
As Burt mentioned, reviewing this commit by commit is easier because some of them are just cleanup/move files/minor changes

burtbeckwith:
would it make more sense to keep partial guides under `/guides`, e.g. `/guides/micronaut-cloud-database-common` containing the shared src and asciidoc includes, with "real" guides `/guides/micronaut-cloud-database-aws`, `/guides/micronaut-cloud-database-azure`, etc.?

the `metadata.json` file in `/guides/micronaut-cloud-database-common` could have a new optional attribute `"partial": true` to indicate that it shouldn't be rendered as a guide, and common attributes could be kept there to reduce duplication

I haven't yet implemented sharing java/groovy/kotlin source or resources between these guides, but it would be good to keep everything common under `/guides/micronaut-cloud-database-core/java`, `/guides/micronaut-cloud-database-core/groovy`, `/guides/micronaut-cloud-database-core/src`, etc. and adding an attribute to the `source:`, `test:`, `resource:`, and `testResource:` macros indicating the source directory

then instead of the new `partial:` syntax we could use the existing `common:` macro with the same source directory attribute

one further option would be to add an attribute in `metadata.json` to indicate the location of a common guide directory and leave the `source:`, `test:`, `resource:`, and `testResource:` as they currently are, but resolve source includes from the common guide directory, similar to the current shared `src` directory logic

sdelamo:
I think we need two things:

- The ability to tell that a guide folder is the `initial` code of a guide. 
- Or a guide folder is the `base` code of a Guide. 

The idea is that if you specify `initial` the reader would the option to download a starter project with some code. If you specify `base` , you are sharing code between guides but no `initial.zip` is generated. For me `base` is what you were refering as `partial`. 

I think what should happend is that if you specify in a guide. e.g. 

`micronaut-microservices-distributed-tracing-zipkin/medata.json`

```json
{
"base": `micronaut-microservices-distributed-tracing`
}
```

It will copy the contents of the base fodler to the `src` folder. Similar what we do currently with the root `src` folder in a guide folder.  If you specify `intial` instead of  `base` we have to generate `initial.zip`. 

As you said, I think we need to ability to say a guide is not published. I would recommend to have something like 

`published: false` in `metadata.json`. It should default to `true`. 

I would recommend handling the `published: false` in a separate PR to make easy to review/evolve the solution. 

@burtbeckwith 

> one further option would be to add an attribute in metadata.json to indicate the location of a common guide directory and leave the source:, test:, resource:, and testResource: as they currently are, but resolve source includes from the common guide directory, similar to the current shared src directory logic

That would be my preferred option.

burtbeckwith:
Closing in favor of https://github.com/micronaut-projects/micronaut-guides/pull/725, and will be reworking support for composed guides in another PR

PR #725, opened 2022-02-01
cleanup from PR #684

burtbeckwith:
I re-did the remaining unmerged work from https://github.com/micronaut-projects/micronaut-guides/pull/684 (minus the changes to support includes from `src/docs/common/partial` which I'm reworking in another PR) so it's more clear what's changing.

As before, look at the individual commits. The first two are non-functional cleanup changes. The 3rd involves passing the Gradle `projectDir` to the Groovy helper classes to simplify the relative paths and make it easier to include from one guide directory to another


sdelamo (approved):

sdelamo:
@burtbeckwith can you make the build pass and merge it once it is green?

PR #728, opened 2022-02-02
Add Micronaut OpenAPI generator server guide

andriy-dmytruk:


guides/micronaut-openapi-generator-server/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+  "asciidoctor": "micronaut-openapi-generator-server.adoc",
+  "slug": "micronaut-openapi-generator-server",
+  "title": "Use OpenAPI Definition to Generate a Micronaut Server",
+  "intro": "Learn how to write an OpenAPI definition, use it to generate a server template for Micronaut framework, and get it all to work",
+  "authors": ["Andriy Dmytruk"],
+  "tags": ["openapi", "server"],
+  "category": "Getting Started",
+  "languages": ["java"],
+  "buildTools": ["maven", "gradle"],
```
burtbeckwith: delete this line, no need to specify `buildTools` if you're using both Gradle and Maven

guides/micronaut-openapi-generator-server/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+  "publicationDate": "2022-02-01",
```
burtbeckwith: be sure to update the date so it's close to the date when it's published on the live site

guides/micronaut-openapi-generator-server/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+  "asciidoctor": "micronaut-openapi-generator-server.adoc",
+  "slug": "micronaut-openapi-generator-server",
+  "title": "Use OpenAPI Definition to Generate a Micronaut Server",
+  "intro": "Learn how to write an OpenAPI definition, use it to generate a server template for Micronaut framework, and get it all to work",
+  "authors": ["Andriy Dmytruk"],
+  "tags": ["openapi", "server"],
+  "category": "Getting Started",
+  "languages": ["java"],
```
burtbeckwith, line 9: are you planning on adding Kotlin and Groovy source?

guides/micronaut-openapi-generator-server/java/src/main/resources/application.yml
```diff
@@ -0,0 +1,18 @@
+context-path: "/"
+
+micronaut:
+    application:
```
burtbeckwith: indent with 2 spaces for consistency

guides/micronaut-openapi-generator-server/java/src/main/resources/application.yml
```diff
@@ -0,0 +1,18 @@
...
+        name: openapi-micronaut
+    server:
+        port: 8080
```
burtbeckwith: no need to specify port 8080, it's the default

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/library/Application.java
```diff
@@ -0,0 +1,9 @@
+package example.micronaut.library;
+
+import io.micronaut.runtime.Micronaut;
+
+public class Application {
```
burtbeckwith: delete this class. during the build, an application is created and your source code is copied into that project. if you need to replace an auto-generated class like this because you've changed something, include it in the guide so it overwrites the generated file, otherwise there's no need

guides/micronaut-openapi-generator-server/metadata.json
burtbeckwith: nvm, only Java makes sense because the code is generated

buildSrc/src/main/java/io/micronaut/guides/feature/SwaggerAnnotations.java
```diff
@@ -0,0 +1,12 @@
+package io.micronaut.guides.feature;
+
+import jakarta.inject.Singleton;
+
+import static io.micronaut.starter.build.dependencies.Scope.TEST;
```
burtbeckwith: delete this unused import

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+
+common:completesolution.adoc[]
+
+== Installing OpenAPI Generator
+
+To use OpenAPI Generator we need to install OpenAPI Generator CLI.
+
+We will use the jar option. Download it from
+https://repo1.maven.org/maven2/org/openapitools/openapi-generator-cli/5.4.0/openapi-generator-cli-5.4.0.jar
+and store it in the folder that you want to use for this project.
```
burtbeckwith: use "directory" instead of "folder" (10 instances)

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+java -jar openapi-generator-cli-5.4.0.jar help
+----
+
+It should provide a description and a list of commands.
+
+[NOTE]
+===============================
+All the options for installation are given at the https://openapi-generator.tech/docs/installation[OpenAPI Generator website].
+
+If you installed the generator with a packet manager or bash launcher script, simply run
```
burtbeckwith: packet -> package

burtbeckwith: maybe show (or just add a doc link) how to install with a package manager or bash script?

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+
+//NOTE: You can get the complete version of the config file
+//that we are about to create here:
+//link:{sourceDir}@sourceDir@/library-definition.yaml[library-definition.yaml].
+//However, we do recommend you to look through each of its sections for better understanding.
+
+Openapi generator supports `.yaml` and `.json` file formats for the definition file.
+We will use YAML due its simplicity and human readability.
+
+In the folder to which you downloaded the OpenAPI generator CLI create a file named `library-definition.yaml` and open it with your
```
burtbeckwith: "to which" -> "where", also add a comma after CLI

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+  title: Library
+  license:
+    name: Apache-2.0
+    url: "https://www.apache.org/licenses/LICENSE-2.0.html"
+tags: # <3>
+  - name: books
+    description: Search for books and add new ones
+----
+<1> The version that will be used for parsing.
+<2> The info object contains general information about the api to its reader.
```
burtbeckwith: api -> API

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+<3> The `search` operation id will be used as method name for the given path.
+<4> We define 2 parameters of type string that user should supply in query.
+<5> Validation can be used on parameters. In this case book name must contain at least 3 characters.
+<6> The `responses` object describes the response codes that can be produced.
+It also defines the structure of body if any.
+<7> In case of correct request we define the body to contain a list of `BookInfo` objects. The schema for the book info
+object will be defined later in `components/schemas` section of definition.
+<8> The `"400"` status code will be produced by Micronaut in case of bad request, like an incorrect type supplied or
+failed validation. Even though Micronaut handles it automatically and no implementation is needed on our side, we add it
+for a complete api specification.
```
burtbeckwith: api -> API

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+
+NOTE: You can read more about parameter descriptions in the
+link:https://swagger.io/docs/specification/describing-parameters/["Describing Parameters" OpenAPI guide].
+All the available types and their validations are described in
+link:https://swagger.io/docs/specification/data-models/data-types/["Data Models (Schemas)" OpenAPI guide].
+
+We will define another path with `POST` operation similarly to the way we created the first one:
```
burtbeckwith: "similarly to the way" -> "like"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+        "400":
+          description: Bad Request
+----
+<1> We define the `POST` method for the `/add` path, and add the same tag `books` to it.
+<2> We specify that a body is required and what are the supported content-types for it.
+(in this case only `application/json`, but multiple can be allowed).
+<3> We write that `BookInfo` object is required to be in the request body.
+We reference the same `BookInfo` schema that we will define next.
+
+NOTE: If you want to read more about body definitions, you can read the
```
burtbeckwith: "If you want to" -> "To", "you can read" -> "see"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+        author: {type: string, minLength: 3}
+        ISBN: {type: string, pattern: "[0-9]{13}"}
+      required: ["name", "availability"]
+    BookAvailability: # <4>
+      type: string
+      enum: ["available", "not available", "reserved"]
+
+----
+<1> We define the `BookInfo` schema inside then `components/schemas` section.
+From this schema a java class will be generated with the same `BookInfo` classname.
```
burtbeckwith: java -> Java, "classname" -> "class name"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+<2> We define all the properties of `BookInfo`, including required validation on them. An abbreviated form is used
+for some YAML lists and dictionaries to reduce the number of rows and simplify readability.
+<3> We reference another scheme to be used as a property.
```
burtbeckwith: scheme -> schema (multiple instances)

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+NOTE: You can read more about writing schemas in the
+link:https://swagger.io/docs/specification/data-models/["Data Models (Schemas)" OpenAPI guide].
+
+Save the file and proceed to the next part of the guide.
+//Altogether it should look like this: link:@sourceDir@/library-definition.yaml[library-definition.yaml].
+
+== Generating server API from template
+
+Now we will generate server API files from our definition.
+Open the terminal in the same folder as `libaray-definition.yaml` file and run the following command:
```
burtbeckwith: libaray -> library

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+----
+<1> Specify that we will use java Micronaut server generator.
+<2> Specify our OpenAPI definition file as `library-definition.yaml` which we just created.
+<3> Specify the output directory to be the current directory (`./`).
+You can specify it to be a different one if you want (e.g. `library-server`).
+<4> We provide generator-specific properties starting with `-p`.
+We want all the controllers to be generated in the `example.micronaut.library.controller` package.
+<5> We want all the models (data models, like `BookInfo`) to be in `example.micronaut.library.model` package.
+<6> We want to use @build@ as build tool. The supported values are `gradle`, `maven` and `all`.
+If nothing is specified, both maven and gradle files are generated.
```
burtbeckwith: maven -> Maven, gradle -> Gradle

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+----
+java -jar openapi-generator-cli-5.4.0.jar config-help \
+    -g java-micronaut-server
+----
+=========
+
+[NOTE]
+=========
+If you plan to change the definition file and regenerate files, consider setting the `-p generateControllerAsAbstract=true`
+parameter (We don't recommend doing it during this guide, though). In this case an abstract class will be generated
```
burtbeckwith: We -> we

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+for the API, while all the logic needs to be implemented in a different class (that inherits the API abstract class).
```
burtbeckwith: inherits -> extends

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+This way your changes won't be overwritten by generation, but API will be updated.
```
burtbeckwith: API -> the API

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+=========
+
+After running OpenAPI generator CLI should output information about generated files.
```
burtbeckwith: "generator CLI" -> "generator, the CLI"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+│           └── example/micronaut/library # <6>
+│               ├── controller
+│               │   └── BooksControllerTest.java
+│               └── model
+│                   ├── BookAvailabilityTest.java
+│                   └── BookInfoTest.java
+├── README.md
+└── ...
+----
+<1> The `docs/` folder contains automatically generated markdown files with documentation about your API.
```
burtbeckwith: markdown -> Markdown

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+<2> The `Application.java` will start the Micronaut server with all the detected Controllers.
```
burtbeckwith: "The `Application.java` will start" -> "Starts", remove "all"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+<3> The `BooksController.java` is generated based on paths with `books` tag. It is generated in the package we specified
```
burtbeckwith: "The `BooksController.java`" -> "`BooksController`"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+If you look inside the generated `BookInfo.java` file, you can see the class that was generated with all the parameters
+based on our definition. Notice that the constructor signature has 2 parameters, which were defined as `required` in the
+YAML definition file:
+
+[source,java]
+----
+    public BookInfo(String name, BookAvailability availability);
+----
+
+Along with that it has getters and setters for parameters and jackson serialization annotation.
```
burtbeckwith: jackson -> Jackson, annotation -> annotations

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+
+To add the required functionality to the server we will first create a service to manage books in our library.
+Create file `src/main/java/example/micronaut/library/BookService.java` with the following content:
```
burtbeckwith: "Create file `src/main/java/example/micronaut/library/BookService.java`" -> "Create `BookService`"

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/library/BookService.java
```diff
@@ -0,0 +1,33 @@
...
+import jakarta.inject.Singleton;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Singleton // <1>
+public class BookService {
+    private static final List<BookInfo> BOOKS = new ArrayList<>(Arrays.asList(
+            new BookInfo("Alice's Advantures in Wonderland", BookAvailability.AVAILABLE)
```
burtbeckwith: Advantures -> Adventures

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/library/BookService.java
```diff
@@ -0,0 +1,33 @@
...
+                    .author("Lewis Carroll"),
+            new BookInfo("The Hitchhiker's Guide to the Galaxy", BookAvailability.RESERVED)
+                    .author("Douglas Adams"),
+            new BookInfo("Java Guide for Beginners", BookAvailability.AVAILABLE)
+    ));
+
+    public void addBook(BookInfo bookInfo) {
+        // Add book to the list
```
burtbeckwith: don't comment the obvious

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+
+source:library/BookService[]
+callout:singleton[1]
+
+Now open the `BooksController.java`. The class `BooksController` should have 2 methods named the same
```
burtbeckwith: "Now open the `BooksController.java`. The class `BooksController` should have 2" -> "Now open `BooksController`; the class should have two"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+as the operations we created in the definition file. The method have Micronaut framework annotations describing the
```
burtbeckwith: method -> methods

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+required API. We will now write their bodies.
+
+First, import and inject a `BookService` instance as a class attribute:
```
burtbeckwith: "import and inject a `BookService` instance as a class attribute:" -> "inject `BookService`:"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+source:library/controller/BooksController[tag=import]
+source:library/controller/BooksController[tag=inject,indent=0]
+<1> Micronaut will create an instance of the class and inject it here.
```
burtbeckwith: "Micronaut will create an instance of the class and inject it here" -> "Micronaut will inject the singleton instance of the bean here"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+
+Next, we will add an implementation of the `search` method:
+
+source:library/controller/BooksController[tag=search,indent=0]
+<1> Return a Reactor publisher, that will return the result of search method.
+
+And finally, we will implement the `addBook` method:
+source:library/controller/BooksController[tag=search,indent=0]
+<1> Call the desired function and return empty string, that to send an empty body in response.
```
burtbeckwith: "Call the desired function and return empty string, that to send an empty body in response." -> "Call the function and return an empty string to create a response with an empty body."

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+[source,bash]
+----
+TP/1.1 200 OK
+date: Tue, 1 Feb 2022 00:01:57 GMT
+Content-Type: application/json
+content-length: 0
+connection: keep-alive
+----
+
+== Testing Application
```
burtbeckwith: "Testing Application" -> "Testing the Application"

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/library/controller/BooksController.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.validation.Valid;
+import javax.validation.constraints.*;
+import io.swagger.annotations.*;
+
+@Generated(value="org.openapitools.codegen.languages.JavaMicronautServerCodegen", date="2022-01-31T13:21:29.860-05:00[America/Toronto]")
+@Controller("${context-path}")
+public class BooksController {
+
+    //tag::inject[]
+    @Inject // <1>
```
burtbeckwith: typically @Inject and a field like this is used for dependency injection in tests, but for application code it's best to use constructor injection with a final field to make it immutable:

```
    private final BookService bookService;

    public BooksController(BookService bookService) { // <1>
        this.bookService = bookService;
    }
```

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+source:library/controller/BooksController[tag=inject,indent=0]
+<1> Micronaut will create an instance of the class and inject it here.
+
+Next, we will add an implementation of the `search` method:
+
+source:library/controller/BooksController[tag=search,indent=0]
+<1> Return a Reactor publisher, that will return the result of search method.
+
+And finally, we will implement the `addBook` method:
+source:library/controller/BooksController[tag=search,indent=0]
```
burtbeckwith: "tag=search" -> "tag=addBook"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+Content-Type: application/json
+content-length: 0
+connection: keep-alive
+----
+
+== Testing Application
+
+As we have noticed previously, some files were generated as templates for tests.
+
+We will modify the `src/test/java/example/micronaut/library/controllerBooksControllerTest` to test our paths.
```
burtbeckwith: "the `src/test/java/example/micronaut/library/controllerBooksControllerTest`" -> "`BooksControllerTest`"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+For path `/search` 2 methods were generated: one for testing the `search` method itself, and another for testing
```
burtbeckwith: spell out numbers less than 10 (i.e. 2 throughout), use digits for numbers 10+

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+We will delete the `searchMethodTest()` and modify the `searchClientApiTest()` method to perform a basic check.
+All the modifications compared to the generated version are labeled here:
+
+test:library/controller/BooksControllerTest[tag=search,indent=0]
+<1> Remove the `@Disabled` annotation, so that the test would run.
+<2> Specify the value of query the parameter.
+<3> Store the response body in a variable.
+<4> Add assertion: we expect to get 2 books containing `"Guide"` as response.
+
+We will also add a test to the `/add` path:
```
burtbeckwith: "We will also add a test to the `/add` path:" -> "We will also update the generated `addBookClientApiTest` for the `/add` path:"

guides/micronaut-openapi-generator-server/java/src/test/java/example/micronaut/library/controller/BooksControllerTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @Inject
+    BooksController controller;
+
+    /**
+     * This test is used to check that the api available to client through
+     * '/add' to the features of addBook() works as desired.
+     */
+    //tag::addBook[]
+    @Test // <1>
+    void addBookClientApiTest() throws IOException {
```
burtbeckwith: remove `throws IOException`

guides/micronaut-openapi-generator-server/java/src/test/java/example/micronaut/library/controller/BooksControllerTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+    //end::addBook[]
+
+    /**
+     * This test is used to check that the api available to client through
+     * '/search' to the features of search() works as desired.
+     */
+    //tag::search[]
+    @Test // <1>
+    void searchClientApiTest() throws IOException {
```
burtbeckwith: remove `throws IOException`

guides/micronaut-openapi-generator-server/java/src/test/java/example/micronaut/library/controller/BooksControllerTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Inject
+    @Client("${context-path}")
+    HttpClient client;
+
+    @Inject
+    BooksController controller;
+
+    /**
+     * This test is used to check that the api available to client through
```
burtbeckwith: api -> API (same for other test)

guides/micronaut-openapi-generator-server/java/src/test/java/example/micronaut/library/controller/BooksControllerTest.java
```diff
@@ -0,0 +1,96 @@
+package example.micronaut.library.controller;
+
+import example.micronaut.library.model.BookAvailability;
```
burtbeckwith: I had to manually add an import for `BookAvailability`, it wasn't in the generated code - is this a bug?

guides/micronaut-openapi-generator-server/java/src/test/java/example/micronaut/library/controller/BooksControllerTest.java
```diff
@@ -0,0 +1,96 @@
...
+                .accept("applicaton/json");
+        request.getParameters()
+                .add("book-name", "Guide"); // <2>
+
+        // when
+        HttpResponse<?> response = client.toBlocking().exchange(request, Argument.of(List.class, BookInfo.class));
+        List<BookInfo> body = (List<BookInfo>) response.body(); // <3>
+
+        // then
+        Assertions.assertEquals(HttpStatus.OK, response.status());
```
burtbeckwith: My preference for JUnit tests is to use static imports for `Assertions` methods - is this possible with the generated code?

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+The tests should both run successfully.
+
+common:graal-with-plugins.adoc[]
+
+== Next steps
+
+=== Add Security
+
+We could have defined our security requirements by adding a security scheme to the `library-definition.yaml` file.
+For example, we will add basic authentication:
```
burtbeckwith: "basic" -> "HTTP Basic"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+        - MyBasicAuth: [] # <2>
+components:
+  schemas:
+    # ... #
+  securitySchemes:
+    MyBasicAuth: # <1>
+      type: http
+      scheme: basic
+----
+<1> Define a security scheme inside the `components/securitySchemes`. We want to use basic auth for authentication.
```
burtbeckwith: basic -> Basic

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+<2> Add the scheme to the paths that you want to secure. In this case we want to restrict access to
+adding books into our library.
+
+NOTE: You can read more about describing various authentication in the
+link:https://swagger.io/docs/specification/authentication/["Authentication and Authorization" OpenAPI guide].
+
+The generator will then annotate such endpoints with Micronaut`s
```
burtbeckwith: "Micronaut`s" -> "the"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+----
+
+You will then need to implement an
+link:https://micronaut-projects.github.io/micronaut-security/latest/guide/#authenticationProviders[AuthenticationProvider]
+that satisfies your needs. If you want to finish implementing the basic authentication, continue to the
+link:https://guides.micronaut.io/latest/micronaut-security-basicauth.html[Micronaut Basic Auth guide] and replicate
+steps to create the `AuthenticationProvider` and appropriate tests.
+
+NOTE: You can also read link:https://micronaut-projects.github.io/micronaut-security/latest/guide/[Micronaut Security documentation]
+or link:https://micronaut.io/guides[Micronaut guides] about security to learn more about all
```
burtbeckwith: delete "all"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+the supported Authorization strategies.
+
+=== Generate Micronaut Client
+
+You can generate Micronaut Client based on the same `library-definition.yaml` file.
```
burtbeckwith: "Micronaut Client" -> "a Micronaut client"

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,497 @@
...
+You can generate documentation in html file inside the `html-docs/` folder by running
+[source,bash]
+----
+java -jar openapi-generator-cli-5.4.0.jar generate \
+    -g html2 \
+    -i library-definition.yaml \
+    -o html-docs
+----
+
+common:helpWithMicronaut.adoc[]
```
burtbeckwith: Delete helpWithMicronaut.adoc, that's only for guides written by Object Computing

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
andriy-dmytruk: Added short description for Homebrew and bash script.
Not sure if that is needed, as the link above takes to all the options for installation (updated its name, too).

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
andriy-dmytruk: Removed the subordinate clause completely ("similarly to the way we created the first one" -> "")

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
andriy-dmytruk: "the OpenAPI generator CLI" is the thing that we run and that outputs information, so probably no comma. Not sure if there is a way to make the sentence more clear, though.

burtbeckwith: How about "After running, the OpenAPI generator CLI will output information about generated files."

guides/micronaut-openapi-generator-server/metadata.json
andriy-dmytruk: Also I am not sure what category this should be put in. "Getting started" might not be the right one.

guides/micronaut-openapi-generator-server/java/.openapi-generator-ignore
```diff
@@ -0,0 +1,23 @@
+# OpenAPI Generator Ignore
```
sdelamo: Should we have this file as part of the complete solution? Or should we delete it?

guides/micronaut-openapi-generator-server/java/.openapi-generator/FILES
```diff
@@ -0,0 +1,22 @@
+.gitignore
```
sdelamo: Should we have this file as part of the complete solution? Or should we delete it?

buildSrc/src/main/resources/pom.xml
```diff
@@ -145,5 +145,12 @@
             <version>4.1.1</version>
         </dependency>
 
+        <!-- Swagger Annotations -->
+        <dependency>
+            <groupId>io.swagger</groupId>
```
sdelamo, line 135: When you add the open api feature in Micronaut CLI or Micronaut Launch we add the following depenency: 

https://github.com/micronaut-projects/micronaut-starter/blob/3.3.x/starter-core/src/main/java/io/micronaut/starter/feature/other/OpenApi.java#L56

`io.swagger.core.v3:swagger-annotations`

Here you are using `io.swagger:swagger-annotations` instead. Which one should we use in the guide and in starter?

buildSrc/src/main/java/io/micronaut/guides/feature/SwaggerAnnotations.java
```diff
@@ -0,0 +1,10 @@
+package io.micronaut.guides.feature;
+
+import jakarta.inject.Singleton;
+
+@Singleton
+public class SwaggerAnnotations extends AbstractFeature {
+    public SwaggerAnnotations() {
+        super("swagger-annotations", "swagger-annotations");
```
sdelamo, line 8: I think it would be worth extracting a swagger feature from the OpenAPI feature. 

https://github.com/micronaut-projects/micronaut-starter/blob/3.3.x/starter-core/src/main/java/io/micronaut/starter/feature/other/OpenApi.java

guides/micronaut-openapi-generator-server/java/.openapi-generator/VERSION
```diff
@@ -0,0 +1 @@
+5.4.0-SNAPSHOT
```
sdelamo: Should we have this file as part of the complete solution? Or should we delete it?

guides/micronaut-openapi-generator-server/java/README.md
```diff
@@ -0,0 +1,23 @@
+# openapi-micronaut
```
sdelamo: Should we have this file as part of the complete solution? Or should we delete it?

guides/micronaut-openapi-generator-server/java/docs/controllers/BooksController.md
```diff
@@ -0,0 +1,52 @@
+# BooksController
```
sdelamo: Should we have this file as part of the complete solution? Or should we delete it?

guides/micronaut-openapi-generator-server/java/docs/models/BookAvailability.md
```diff
@@ -0,0 +1,17 @@
+
+
+# BookAvailability
```
sdelamo: Should we have this file as part of the complete solution? Or should we delete it?

guides/micronaut-openapi-generator-server/java/docs/models/BookInfo.md
```diff
@@ -0,0 +1,22 @@
+
+
+# BookInfo
```
sdelamo: Should we have this file as part of the complete solution? Or should we delete it?

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/BookService.java
```diff
@@ -0,0 +1,32 @@
...
+import example.micronaut.model.BookAvailability;
+import example.micronaut.model.BookInfo;
+import jakarta.inject.Singleton;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Singleton // <1>
+public class BookService {
```
sdelamo, line 13: I think we should save the `BookInfo` in a MySQL database with Flyway migrations and using Micronaut data JDBC. 

I think this will involve creating a `JdbcRepository` a `MappedEntity` and a flyway migration.

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/BookService.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final List<BookInfo> BOOKS = new ArrayList<>(Arrays.asList(
+            new BookInfo("Alice's Adventures in Wonderland", BookAvailability.AVAILABLE)
+                    .author("Lewis Carroll"),
+            new BookInfo("The Hitchhiker's Guide to the Galaxy", BookAvailability.RESERVED)
+                    .author("Douglas Adams"),
+            new BookInfo("Java Guide for Beginners", BookAvailability.AVAILABLE)
+    ));
+
+    public void addBook(BookInfo bookInfo) {
+        BOOKS.add(bookInfo);
```
sdelamo, line 23: Should the addition of a book be synchronized?

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/controller/BooksController.java
```diff
@@ -0,0 +1,102 @@
...
+    @Produces(value = {})
+    @Consumes(value = {"application/json"})
+    @Secured(SecurityRule.IS_ANONYMOUS)
+    //tag::addBook[]
+    public Mono<Object> addBook(
+        @Body @NotNull @Valid BookInfo bookInfo
+    ) {
+        return Mono.fromCallable(() -> {
+            bookService.addBook(bookInfo);
+            return ""; // <1>
```
sdelamo, line 69: this feels a bit weird to return an empty string.

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/model/BookAvailability.java
```diff
@@ -0,0 +1,62 @@
+/*
```
sdelamo: I think we should attempt to generate this file as part of the guide generation.

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/model/BookInfo.java
```diff
@@ -0,0 +1,195 @@
+/*
```
sdelamo: I think we should attempt to generate this file as part of the guide generation.

guides/micronaut-openapi-generator-server/metadata.json
```diff
@@ -0,0 +1,17 @@
...
+  "publicationDate": "2022-02-07",
+  "apps": [
+    {
+      "name": "default",
+      "features": ["netty-server", "http-client", "security", "reactor", "swagger-annotations"]
```
sdelamo: I don't think you need to add `netty-server` or `http-client` features

sdelamo (changes requested):
Great work @andriy-dmytruk 

However, I think like us to expore if we could generate the guide files `BookAvailability` `BookInfo`  automatically given the Open API yaml definition.  

They have a core dependency or a gradle plugin which we may be able to leverage:

https://github.com/OpenAPITools/openapi-generator#12---artifacts-on-maven-central

Generating the files automatically will allow us to keep the guide almost up to date when new versions of OpenAPI Generator is released. 

Moreover, I have done sever commits @andriy-dmytruk Let me know if you have concerns with any of those commits.

guides/micronaut-openapi-generator-server/metadata.json
sdelamo: Add a new category to the enum named "API"

guides/micronaut-openapi-generator-server/java/docs/controllers/BooksController.md
andriy-dmytruk: This could not be generated at all by setting `-D modelDocs=false -D apiDocs=false`, but by default is generated.
I would keep so that the directory structure would be the same as user would get if they generate it.

**The same is true for all `/docs` files.**

guides/micronaut-openapi-generator-server/java/.openapi-generator/FILES
andriy-dmytruk: This and the version file is not used by the generator in any way. It seems to be some kind of debug file or something to know what files were generated and by version in case user needs to know.
It can be skipped by user by setting the `-D generateMetadata=false` option.
We can delete it. But if someone will try running the generation themselves, they will have it.

buildSrc/src/main/resources/pom.xml
andriy-dmytruk: It seems like `io.swagger.core.v3:swagger-annotations` is the newer version. Thanks.
I will have to change this in the generator.

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -170,7 +170,14 @@ If you are using Windows command prompt, run:
...
 ----
-java -jar openapi-generator-cli-5.4.0.jar generate -g java-micronaut-server -i library-definition.yaml -o ./ -p controllerPackage=example.micronaut.controller -p modelPackage=example.micronaut.model -p build=@build@ -p test=junit
+java -jar openapi-generator-cli-5.4.0.jar generate \
+     -g java-micronaut-server \
+     -i library-definition.yaml  \
+     -o ./
+     -p controllerPackage=example.micronaut.controller  \
+     -p modelPackage=example.micronaut.model  \
+     -p build=@build@  \
+     -p test=junit
```
andriy-dmytruk: Unluckily this won't work on Windows. Windwos Power Shell uses `'` to command continuation to new line, while command prompt uses `^` (I don't know why they did it in such a way, but it seems like it cannot be formatted properly😕).

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/BookService.java
andriy-dmytruk: Maybe we can add it as a next step (like I added some comments about security)?
I don't think it should be required to have MySQL database to complete the guide.

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/controller/BooksController.java
andriy-dmytruk: I agree. But I think returning `null` will result in an error. And it is different if we change the method to return an `HTTPResponse` object - I suppose I will need to change generation to this. Is there another way?

andriy-dmytruk:
@sdelamo I started generating the model files automatically using dependancy to [OpenAPI Generator core](https://mvnrepository.com/artifact/org.openapitools/openapi-generator/5.4.0). Could you take a look at the current implementation and refactor if needed.?

I suppose it is best if `build.gradle` and `pom.xml` files are generated like before, therefore I run the OpenAPI generator before everything. This however has the disadvantage, that `application.yml` cannot be automatically generated as it would be overwritten. But I guess it is part of a larger question of also using OpenAPI generator in starter, and this should be fine for now.

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+        if (!definitionFilePath.isFile()) {
+            throw new GradleException("OpenAPI definition file '" + definitionFilePath.getPath() + "' does not exist")
+        }
+        CodegenConfigurator configurator = new CodegenConfigurator();
+        configurator.setInputSpec(definitionFilePath.getPath());
+        configurator.setGeneratorName(config.generatorName)
+        configurator.setOutputDir(destination.getPath())
+
+        if (testFramework == SPOCK) {
+            configurator.addAdditionalProperty("test", "spock")
```
sdelamo: please extract constants for "test" and "spock"

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+        } else if (testFramework == JUNIT) {
+            configurator.addAdditionalProperty("test", "junit")
```
sdelamo: please extract constants for "test" and "junit"

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+        }
+        if (buildTool == GRADLE) {
+            configurator.addAdditionalProperty("build", "gradle")
```
sdelamo: please extract constants for "build" and "gradle"

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+        } else if (buildTool == MAVEN) {
+            configurator.addAdditionalProperty("build", "maven")
```
sdelamo: please extract constants for "build" and "maven"

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+        }
+        configurator.addAdditionalProperty("controllerPackage", destinationPackage + ".controller")
```
sdelamo: please extract constants for "controllerPackage" and ".controller"

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+        configurator.addAdditionalProperty("apiPackage", destinationPackage + ".api")
```
sdelamo: please, extract constants for "apiPackage" and ".api"

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+        configurator.addAdditionalProperty("modelPackage", destinationPackage + ".model")
```
sdelamo: please extract constants for "modelPackage" and ".model"

sdelamo (changes requested):

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/controller/BooksController.java
sdelamo: In a real world API I would probably return 201 created with show uri in the HTTP location header

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
sdelamo: This won't work in MacOS either. It is just to display it better in the guide. I understand our readers will be able to take this command and format it to their OS. 

Do you think we should revert this command and use a one liner?

buildSrc/src/main/resources/pom.xml
sdelamo: Can you update the guide to use `io.swagger.core.v3:swagger-annotations` in the meanwhile?

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+import org.openapitools.codegen.Generator
+import org.openapitools.codegen.GeneratorNotFoundException
+import org.openapitools.codegen.config.CodegenConfigurator
+
+import static io.micronaut.starter.api.TestFramework.JUNIT
+import static io.micronaut.starter.api.TestFramework.SPOCK
+import static io.micronaut.starter.options.BuildTool.GRADLE
+import static io.micronaut.starter.options.BuildTool.MAVEN
+
+@Singleton
```
sdelamo: No need to annotate this as `@Singleton`.

sdelamo (changes requested):

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+        configurator.setInputSpec(definitionFilePath.getPath());
+        configurator.setGeneratorName(config.generatorName)
+        configurator.setOutputDir(destination.getPath())
+
+        if (testFramework == SPOCK) {
+            configurator.addAdditionalProperty("test", "spock")
+        } else if (testFramework == JUNIT) {
+            configurator.addAdditionalProperty("test", "junit")
+        }
+        if (buildTool == GRADLE) {
```
sdelamo: We currently don't generate Guides with the Gradle Kotlin DSL but in the future, we may. There is a convenience method in `BuildTool::isGradle` which should be used instead.

We should replace 

`if (buildTool == GRADLE) {`

with: 

`if (buildTool.isGradle) {`

sdelamo (changes requested):

guides/micronaut-openapi-generator-server/metadata.json
```diff
@@ -0,0 +1,21 @@
...
+  "category": "Getting Started",
+  "languages": ["java"],
+  "publicationDate": "2022-02-07",
+  "apps": [
+    {
+      "name": "default",
+      "features": ["security", "reactor", "swagger-annotations"],
+      "openAPIGeneratorConfig": {
+        "definitionFile": "src/main/resources/library-definition.yaml",
+        "generatorName": "java-micronaut-server"
```
sdelamo: I assume we are always going to use the generator of `java-micronaut-server` in guides. Thus, I think we can default to this.

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -0,0 +1,70 @@
...
+            configurator.addAdditionalProperty("build", "gradle")
+        } else if (buildTool == MAVEN) {
+            configurator.addAdditionalProperty("build", "maven")
+        }
+        configurator.addAdditionalProperty("controllerPackage", destinationPackage + ".controller")
+        configurator.addAdditionalProperty("apiPackage", destinationPackage + ".api")
+        configurator.addAdditionalProperty("modelPackage", destinationPackage + ".model")
+
+        config.properties?.each {
```
sdelamo: in groovy you can do this more idiomatic by using: 

`.each { k, v ->`

sdelamo (changes requested):

guides/micronaut-openapi-generator-server/metadata.json
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/b34e78642671134c75b7cab67d372e8ac40ccc21

sdelamo:
> @sdelamo I started generating the model files automatically using dependancy to [OpenAPI Generator core](https://mvnrepository.com/artifact/org.openapitools/openapi-generator/5.4.0). Could you take a look at the current implementation and refactor if needed.?

It is looking good!

> I suppose it is best if `build.gradle` and `pom.xml` files are generated like before, therefore I run the OpenAPI generator before everything. This however has the disadvantage, that `application.yml` cannot be automatically generated as it would be overwritten. But I guess it is part of a larger question of also using OpenAPI generator in starter, and this should be fine for now.

I have done a commit to remove every file generated by the open api generator but the source files: 

https://github.com/micronaut-projects/micronaut-guides/pull/728/commits/a4825b40dfca6b792080a9ba623509756d9e039c

We have to revisit the Micronaut open api generator to see if we can tweak the application.yml which we are currenlty geneator. I want to understand the Jackson features we are adding by default.

sdelamo:
@andriy-dmytruk 

I've created a PR to 3.3.1 https://github.com/OpenAPITools/openapi-generator/pull/11569

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
andriy-dmytruk: This command is exactly the same to the one in Linux, so we can either remove it (and let people reformat the Linux one to work on their system) or keep the one-liner, so that they can copy it and run without any issues.

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
```diff
@@ -67,4 +60,56 @@ class OpenAPIGenerator {
...
+                                                                         @NonNull TestFramework testFramework,
+                                                                         @NonNull BuildTool buildTool,
+                                                                         @NonNull String destinationPackage) {
+        Map<String, Object> additionalProperties = [:]
+        testProperty(testFramework).ifPresent(value -> additionalProperties.put(TEST, value))
+        buildProperty(buildTool).ifPresent(value -> additionalProperties.put(PROPERTY_BUILD, value))
+        additionalProperties.put(PROPERTY_CONTROLLER_PACKAGE, destinationPackage + PACKAGE_CONTROLLER)
+        additionalProperties.put(PROPERTY_API_PACKAGE, destinationPackage + PACKAGE_API)
+        additionalProperties.put(PROPERTY_MODEL_PACKAGE, destinationPackage + PACKAGE_MODEL)
+        additionalProperties.put(PROPERTY_GENERATOR_NAME, GuideMetadata.OpenAPIGeneratorConfig.GENERATOR_JAVA_MICRONAUT_SERVER)
```
andriy-dmytruk: Should probably be `config.generatorName` instead of `GuideMetadata.OpenAPIGeneratorConfig.GENERATOR_JAVA_MICRONAUT_SERVER`.
There is also `java-micronaut-client` that can be used in the guides, so the property is needed.

guides/micronaut-openapi-generator-server/metadata.json
andriy-dmytruk: Great

guides/micronaut-openapi-generator-server/java/src/test/java/example/micronaut/library/controller/BooksControllerTest.java
sdelamo: I have replaced the Controller test completely. It uses statics imports now

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
sdelamo: Let's keep one command only. We use this format `\` as a separator for readability do in other guides.

buildSrc/src/main/groovy/io/micronaut/guides/OpenAPIGenerator.groovy
sdelamo: > Should probably be config.generatorName instead of GuideMetadata.OpenAPIGeneratorConfig.GENERATOR_JAVA_MICRONAUT_SERVER

You are right. Please, do the change.

guides/micronaut-openapi-generator-server/java/src/test/java/example/micronaut/library/controller/BooksControllerTest.java
andriy-dmytruk: Yes, it will have to be changed in the generator. Thanks.

buildSrc/src/main/resources/pom.xml
andriy-dmytruk: We cannot change to `io.swagger.core.v3:swagger-annotations` now.
They have a lot of differences with the current version, and most imports do not work, e.g.:
`import io.swagger.annotations.ApiModel` -> `import io.swagger.v3.oas.annotations.media.Schema`.
Since it is in the model files, we cannot update the imports.

This will have to be changed in the generator. After that we can update the guide to not use `swagger-annotations` as a feature, but simply `openapi`.

PR #743, opened 2022-02-05
delete external macro

sdelamo:
I would prefer to just have one way of using common snippets.

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
+++ b/buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
@@ -21,14 +21,15 @@ import static io.micronaut.starter.api.TestFramework.SPOCK
 class GuideAsciidocGenerator {
 
     private static final String INCLUDE_COMMONDIR = 'common:'
+    private static final String COMMON_PATH = 'src/docs/common/snippets/common-'
     private static final String CALLOUT = 'callout:'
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
```diff
@@ -166,9 +165,7 @@ class GuideAsciidocGenerator {
             if (rawLine.startsWith(CALLOUT) && rawLine.endsWith(']')) {
                 rawLines << callout(rawLine, projectDir)
             } else if (rawLine.startsWith(INCLUDE_COMMONDIR) && rawLine.endsWith('[]')) {
-                include rawLine, rawLines, projectDir, true
-            } else if (rawLine.startsWith(EXTERNAL) && rawLine.endsWith(']')) {
-                include rawLine, rawLines, projectDir, false
+                include(rawLine, rawLines, projectDir)
```
sdelamo, line 170: I think we should use parenthesis in general when invoking methods. Not sure what is your opinion about this.

burtbeckwith:
My preference is to keep snippets separate from partial guide parts. A snippet can be used from nearly any guide, but partial guides only make sense as part of a "base" guide that N (2 or more, but likely not many) build from. Putting the files in N+1 guides subdirectories keeps everything logically connected.

At a minimum, I'd like to go back to my initial suggestion where we keep partial guide adoc files under `src/docs/common/partial` (or some other subdirectory name)

PR #749, opened 2022-02-06
design: increase width of guides starting at fourth section

sdelamo:
Screenshot 2022-02-06 at 08 00 56


burtbeckwith:
looks great

burtbeckwith (approved):

PR #759, opened 2022-02-08
derive value of asciidoctor and slug

burtbeckwith:
no need to specify the `asciidoctor` or `slug` attributes in metadata.json since `slug` is always the same as the guide directory name and `asciidoctor` is always (for guides where publish=true) the same as the guide directory name plus "adoc" extension


sdelamo (approved):
This is a nice convention

PR #769, opened 2022-02-11
Try run tests in parallel

graemerocher:

PR #780, opened 2022-02-18
First pass at re-modelling the build

timyates:
Still needs a lot of work, we rely a lot on finalizedBy

But this is a good first step, and the no-modification UP-TO-DATE handling is much
better than before


timyates:
Not sure whether to continue with the improvements here, or whether to raise another issue for a further round of polish...

This change basically splits all of the code projects and document generation out per-project and sets the correct inputs.

This means that for a change to a single guide, we only rebuild that guide and a small run of the asciidoc task

I moved all of this logic into a Plugin, but reused the generators to handle it.

Because we build the code projects per-guide now, they moved into their own folders (so caching becomes a possibility later on), this required a change to the asciidoc macros we have written to take this into account, but I *think* it's generating exactly the same as before

timyates:
Ahhhhh... I need to add the JDK version logic into the plugin 😿

settings.gradle
```diff
@@ -1 +1,12 @@
+pluginManagement {
+    repositories {
+        gradlePluginPortal()
+        mavenCentral()
+    }
+}
+
+plugins {
+    id 'io.micronaut.build.shared.settings' version '5.1.1'
```
burtbeckwith: should use 5.2.1

timyates: Fixed!  Thanks Burt 😎👍

timyates:
Hmmm.. The failure doesn't fail locally, and `A build scan was not published as you have not authenticated with server 'ge.micronaut.io'.`

Anyone know how to hook up GE?

graemerocher:
Perhaps it could build the matrix dynamically like in https://github.com/micronaut-projects/micronaut-guides/pull/769/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR32-R33

PR #783, opened 2022-02-21
Common create app features

shingjo:
Updated guides to use `common:create-app-features.adoc[]`

Note: Have not done all the guides, yet. These were the once that didn't require other changes to the text.


guides/micronaut-http-client/metadata.json
```diff
@@ -8,7 +8,7 @@
   "apps": [
     {
       "name": "default",
-      "features": ["graalvm", "reactor"]
+      "features": ["reactor", "graalvm"]
```
burtbeckwith, line 11: why change the order?

shingjo: So that `graalvm` comes at the end of the create app command, since it is 'optional'.
```
mn create-app example.micronaut.micronautguide \
    --features=reactor,graalvm \
    --build=gradle --lang=java
```
If a developer was creating something similar to the guide and didn't want GraalVM it was easier to remove.

Debated whether or not to move the `--feature` line to the bottom.

burtbeckwith: moving features to the end makes sense

burtbeckwith (approved):

sdelamo (approved):
Thanks @shingjo

PR #793, opened 2022-02-22
add Groovy and Kotlin source for micronaut-oauth2-oidc-google

burtbeckwith:


guides/micronaut-oauth2-oidc-google/groovy/src/main/groovy/example/micronaut/HostedDomainClaimValidator.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+    HostedDomainClaimValidator(ApplicationConfiguration applicationConfiguration) {
+        this.hostedDomain = applicationConfiguration.hostedDomain
+    }
+
+    @Override
+    boolean validate(OpenIdClaims claims,
+                     OauthClientConfiguration clientConfiguration,
+                     OpenIdProviderMetadata providerMetadata) {
+        def hd = claims.get(HOSTED_DOMAIN_CLAIM)
```
sdelamo, line 28: can you use the type instead of `def`?

sdelamo (changes requested):

guides/micronaut-oauth2-oidc-google/groovy/src/main/groovy/example/micronaut/HostedDomainClaimValidator.groovy
burtbeckwith: it was `Object` in the Java code

PR #796, opened 2022-02-22
add Kotlin source for micronaut-error-handling

burtbeckwith:


burtbeckwith:
@sdelamo BookCreateSpec fails in the Kotlin projects. I spent some time digging but didn't get anywhere - title is null in BookCreatePage when called from line 36. Please take a look, but feel free to delete the Kotlin source if you can't find the problem

PR #803, opened 2022-02-24
add Groovy and Kotlin source for micronaut-data-access-mybatis

burtbeckwith:


sdelamo (approved):
@burtbeckwith you have replaced `Long`, `Integer` with primitives `int`, `long`. 

Avoiding primitives was intentional. I would prefer us to discuss styling changes before creating PRs changing the guide style. I prefer to use Objects instead of primitives.

PR #808, opened 2022-02-25
Groovy and Kotlin source for micronaut-mongodb-synchronous 

burtbeckwith:


guides/micronaut-mongodb-synchronous/java/src/test/java/example/micronaut/MockFruitRepository.java
```diff
@@ -0,0 +1,23 @@
+package example.micronaut;
+
+import io.micronaut.context.annotation.Replaces;
+import io.micronaut.context.annotation.Requires;
+import jakarta.inject.Singleton;
+
+import java.util.Collections;
+import java.util.List;
+
+@Requires(property = "spec.name", value = "FruitControllerValidationTest")
```
sdelamo, line 10: I personally prefer to have this classes in the same test file

sdelamo (approved):

guides/micronaut-mongodb-synchronous/java/src/test/java/example/micronaut/MockFruitRepository.java
burtbeckwith: Me too, and it works well in Java and Groovy. But I've had issues with that in Kotlin. It seems like it's a problem with Micronaut during compilation, but it's hard to tell because the errors don't have much information

PR #811, opened 2022-02-25
You need to add the micronaut-data-hibernate-jpa and micronaut-data-p…

kurosouza:
…rocessor dependencies to build.gradle for it to work

The testGenreCrudOperationTest kept failing until I added this.


burtbeckwith:
@kurosouza We have CI tests for all the guides which would have caught an issue like this. Additionally I added Groovy source recently and found a classpath issue (only affected Maven for me though) and updated the guide. Could you try the guide again and see if this is still an issue?

If it still is, lets us know what OS and version of Java you're using, and the version of Micronaut.

sdelamo:
Closing this for now. @kurosouza could you provide a sample app which reproduces the issue?

PR #812, opened 2022-02-28
Fix callouts in mongo POJO kotlin class

timyates:
The callouts were not working for the Fruit POJO (POKO?) class in the mongo db guides (async and sync)

Also spotted that code changes were not forcing a re-run of the asciidoc task


sdelamo (approved):
Thanks @timyates

PR #828, opened 2022-03-07
Groovy source for micronaut-jpa-hibernate

burtbeckwith:


burtbeckwith:
I added Kotlin source also but hit https://github.com/micronaut-projects/micronaut-data/issues/1368, so I'll do the Kotlin changes in a separate PR

buildSrc/src/main/java/io/micronaut/guides/feature/JakartaPersistenceApi2.java
```diff
@@ -0,0 +1,34 @@
...
+    public String getName() {
+        return "jakarta-javax-persistence";
+    }
+
+    @Override
+    public void apply(GeneratorContext generatorContext) {
+        generatorContext.addDependency(Dependency.builder()
+                .groupId("jakarta.persistence")
+                .artifactId("jakarta.persistence-api")
+                .version("2.2.3")
```
sdelamo, line 26: does not work for 3.0.0? We have feature for that.

buildSrc/src/main/java/io/micronaut/guides/feature/JakartaPersistenceApi2.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static io.micronaut.starter.build.dependencies.Scope.COMPILE;
+
+@Singleton
+public class JakartaPersistenceApi2 implements Feature {
+
+    @NonNull
+    @Override
+    public String getName() {
+        return "jakarta-javax-persistence";
```
sdelamo, line 18: if we need v 2.2.3 I think a better name would be jakarta-persistence-api-v2

sdelamo (approved):

buildSrc/src/main/java/io/micronaut/guides/feature/JakartaPersistenceApi2.java
burtbeckwith: The packages changed in 3.0.0 from `javax.persistence` to `jakarta.persistence`

buildSrc/src/main/java/io/micronaut/guides/feature/JakartaPersistenceApi2.java
burtbeckwith: I doubt this will be used again, it's only needed for hibernate-jpa which currently uses v2.2.3. I assume it'll be updated to v3.0.0 at some point and we can delete this

Issue #851, opened 2022-03-16
Project should publish to Gradle Enterprise


timyates:
We should be publishing runs to Gradle Enterprise, but we do not wish to add our full build plugin

sdelamo:
We have these secrets available: 

- `GRADLE_ENTERPRISE_ACCESS_KEY`

- `GRADLE_ENTERPRISE_CACHE_PASSWORD`

- `GRADLE_ENTERPRISE_CACHE_USERNAME`

we will need to add it to github actions https://github.com/micronaut-projects/micronaut-security/blob/master/.github/workflows/gradle.yml#L45-L47

sdelamo:
@timyates work on the branch of this PR https://github.com/micronaut-projects/micronaut-guides/pull/852/

timyates:
Closed by https://github.com/micronaut-projects/micronaut-guides/pull/852

Issue #850, opened 2022-03-16
run guide tests gradle task should be disabled if the guide does not support the JDK version being used


sdelamo:
For example, if you use JDK 11 the gradle task to execute the tests of the guide `micronaut-java-records` which is only JDK 17 should not be enabled.

timyates:
All the guides are created with

```
java {
    sourceCompatibility = JavaVersion.toVersion("11")
    targetCompatibility = JavaVersion.toVersion("11")
}
```

So will never run on Java 8 🤔

sdelamo:
don't we create them with the JDK being used?

timyates:
This is getting done in https://github.com/micronaut-projects/micronaut-guides/pull/852 as well as the connection to Gradle Enterprise

timyates:
Fixed in #852

timyates:
> don't we create them with the JDK being used?

It doesn't look to be that way...  I guess they should be?  They seems to come in as 11 all the time even if I'm on JDK 8 🤔

That's why I originally did the TODO guide as Java 11 https://github.com/micronaut-projects/micronaut-guides/pull/822#discussion_r824329890

sdelamo:
For each guide we ship a ZIP file wit the complete solution. The idea is that the ZIP is JDK 11 but if you have a minimum requirement of 17 it should be 17.

sdelamo:
Java records already is generated with 17 https://guides.micronaut.io/latest/micronaut-java-records-gradle-java.html see ZIP file there.

timyates:
Cool, so we cannot run the tests on Java 8

timyates:
I'll close this then

PR #852, opened 2022-03-16
ci action with runAllGuideTests

sdelamo:
This PR changes our Github actions to run the gradle task `runAllGuideTests` for every PR. With a build with properly defined inputs and outputs running a new guide should only execute the tests for the new guide. 

Relates to: 

https://github.com/micronaut-projects/micronaut-guides/issues/851

https://github.com/micronaut-projects/micronaut-guides/issues/850


.github/workflows/gradle.yml
```diff
@@ -0,0 +1,36 @@
...
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Run All Guide Tests
+        run: 'export AWS_ACCESS_KEY_ID=XXX;export AWS_SECRET_ACCESS_KEY=YYY;export AWS_REGION=us-east-1;./gradlew runAllGuideTests'
```
timyates: Should these `export` lines be in an `env` block?  And not sure XXX and YYY is correct unless there's some magic that replaces them?

timyates (changes requested):

timyates:
I added Gradle Enterprise configuration and the plugin.  Currently there are no labels, and the cache is not configured.  I'm trying to find out how these are calculated in the settings plugin

timyates:
I have removed Java 8 from the matrix until we decide on https://github.com/micronaut-projects/micronaut-guides/issues/850#issuecomment-1069041374

.github/workflows/gradle.yml
timyates: Moved them

PR #855, opened 2022-03-16
Use a worker for running the script and make cacheable

timyates:
This switches from an Exec task to a Worker action so we can try to parallelise in the future.

Currently they run serially.

This also tees the output of each script to a file, and makes the test running cacheable


timyates:
Waiting for the actions to kick in to check it is quicker on CI

sdelamo (approved):

PR #856, opened 2022-03-17
build: Parallelization of test runners

timyates:
- Use `mustRunAfter` instead of a boolean
- Make Gradle more verbose to triage any errors

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/guides/GuidesPlugin.groovy
+++ b/buildSrc/src/main/groovy/io/micronaut/guides/GuidesPlugin.groovy
@@ -19,7 +19,6 @@ import org.gradle.api.tasks.Copy
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.bundling.Zip
 
-import java.util.concurrent.atomic.AtomicBoolean
 import java.util.function.Predicate
 import java.util.stream.Collectors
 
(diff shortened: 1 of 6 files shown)


timyates:
@sdelamo Results of this:

### Java 11:

- Pre-parallelisation we [took 2h20m](https://ge.micronaut.io/s/o45k2jyywktvy/timeline) to run all the tests in CI
- After this change, we [took 1h 54m](https://ge.micronaut.io/s/niifvebrpscfi/timeline) to run them

### Java 17:

- Pre-parallelisation: [1h 38m](https://ge.micronaut.io/s/jvq3ijek6v35w/timeline)
- Parallelised: [1h 42m](https://ge.micronaut.io/s/y6bujyrvjxklg/timeline)

### Summary

- For Java 11, we saw almost a 20% speed increase.
- For Java 17, the speed didn't significantly change (it was slightly slower, but a single run may not be a true figure)
- A lot of the tests fire off docker containers, and the CI agent only has 2 cores
- There's probably a lot of variability due to resource contention when we use both of these cores for running tests

### Conclusion

This is probably worth doing.  But I suspect there will be variability in the results due to a lack of CPU resources.

I won't be upset if we decide not to make this change, but I think if we don't we should stop using the atomic boolean, and use the `mustRunAfter` method in this PR.  But that can be done once 3.4.x pressure is off

sdelamo (approved):

PR #858, opened 2022-03-18
ci: add snapshot github action

sdelamo:
This runs with a cron job and increments the version in `version.txt` to the next snapshot before running the tests.


timyates (approved):
Nice

image

PR #863, opened 2022-03-22
Add Kotlin version of the GraphQL ToDo guide

timyates:
This adds the Kotlin version of the GraphQL ToDo guide.

I've used ideomatic Kotlin, but found a couple of oddities:

- `AuthorRepository` has to be an abstract class, not an interface or compilation fails
- `CreateToDoDataFetcher` has to be declared as `open` or `@Transactional` will fail to be applied

I also fixed a couple of issues with the Java version (callouts) I spotted whilst doing this

### DIFF
--- a/guides/micronaut-graphql-todo/kotlin/src/main/kotlin/example/micronaut/ToDoRepository.kt
+++ b/guides/micronaut-graphql-todo/kotlin/src/main/kotlin/example/micronaut/ToDoRepository.kt
@@ -5,4 +5,4 @@ import io.micronaut.data.model.query.builder.sql.Dialect.POSTGRES
 import io.micronaut.data.repository.PageableRepository
 
 @JdbcRepository(dialect = POSTGRES) // <1>
-interface ToDoRepository : PageableRepository<ToDo?, Long?> // <2>
+interface ToDoRepository : PageableRepository<ToDo, Long> // <2>
--- a/guides/micronaut-graphql-todo/kotlin/src/main/kotlin/example/micronaut/ToDoRepository.kt
+++ b/guides/micronaut-graphql-todo/kotlin/src/main/kotlin/example/micronaut/ToDoRepository.kt
@@ -1,0 +1,8 @@
+package example.micronaut
+
+import io.micronaut.data.jdbc.annotation.JdbcRepository
+import io.micronaut.data.model.query.builder.sql.Dialect.POSTGRES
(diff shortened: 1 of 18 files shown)


timyates:
Discovered a bug running maven tests if the first test came from the cache.  

Previously, in order that a single task downloaded the Maven wrapper,  tasks were organised like this

```mermaid
graph TD;
    A-- must run after -->B;
    A-- must run after -->C;
    A-- must run after -->D;
```

Each of them runs maven, so only after the first test task completed, did the rest of the tasks run in parallel via the Worker API. This means we only downloaded the maven wrapper once, and did not download it simultaneously in two later concurrent tasks.

However if the first task came from the cache

```mermaid
graph TD;
    A{cached}-- must run after -->B;
    A-- must run after -->C;
    A-- must run after -->D;
```

Then all of the later tasks that were not cached would run simultaneously, and fail because the maven wrapper was an invalid ZIP.

The fix here https://github.com/micronaut-projects/micronaut-guides/pull/863/commits/29ceff9752c7502f40a108289dc3c51e2263c488 adds an AtomicBoolean initially set to `true`, which causes the first task to do work to block the Worker API (with `await()`) until it has completed.  It then sets this global AtomicBoolean to `false`, and the remaining tasks run in parallel

guides/micronaut-graphql-todo/kotlin/src/main/kotlin/example/micronaut/ToDoRepository.kt
```diff
@@ -0,0 +1,8 @@
+package example.micronaut
+
+import io.micronaut.data.jdbc.annotation.JdbcRepository
+import io.micronaut.data.model.query.builder.sql.Dialect.POSTGRES
+import io.micronaut.data.repository.PageableRepository
+
+@JdbcRepository(dialect = POSTGRES) // <1>
+interface ToDoRepository : PageableRepository<ToDo?, Long?> // <2>
```
sdelamo: should this be `<Todo,Long>` instead of `<Todo?,Long?>`?

sdelamo (changes requested):

guides/micronaut-graphql-todo/kotlin/src/main/kotlin/example/micronaut/ToDoRepository.kt
timyates: Done 👍

PR #888, opened 2022-04-13
Add tip about using mysql:oracle docker image for Mac M1 users

wetted:
Add tip about using mysql:oracle docker image for Mac M1 users, since there is no darwin/arm64 version of mysql:8.


src/docs/common/snippets/common-docker-mysql-arm.adoc
```diff
@@ -0,0 +1 @@
+TIP: If you are using macOS on Apple Silicon – e.g. M1, M1 Pro, etc. – Docker might fail to pull a native image for `mysql:8`. In that case substitute `mysql:oracle` and make the corresponding change for the `testcontainers datasource` in `src/test/respources/application-test.yml`.
```
burtbeckwith:
```suggestion
TIP: If you are using macOS on Apple Silicon – e.g. M1, M1 Pro, etc. – Docker might fail to pull an image for `mysql:8`. In that case substitute `mysql:oracle` and make the corresponding change for the `testcontainers datasource` in `src/test/resources/application-test.yml`.
```

sdelamo (approved):

PR #890, opened 2022-04-14
build: Micronaut to 3.4.2

sdelamo:


timyates (approved):
🎉

Once this is merged, I'll get maven added to the data-mongo guides

PR #902, opened 2022-04-25
Composite Pattern

sdelamo:
This is a new guide explaining the Composite Pattern which we use in several places across the framework. Locale Resolution, Host resolution, etc

To build it locally: 

`./gradlew micronautPatternsCompositeBuild`


guides/micronaut-patterns-composite/micronaut-patterns-composite.adoc
```diff
@@ -0,0 +1,108 @@
+common:header-top.adoc[]
+
+common:requirements.adoc[]
+
+common:completesolution.adoc[]
+
+== Composite Pattern
+
+Imagine you want to create an API to resolve a color:
```
alvarosanchez, line 9: The guide is great. But I miss an introduction paragraph. You jump straight to the solution of a problem/need that is worth introducing first, IMO.

alvarosanchez (changes requested):
The guide is great. But I miss an introduction paragraph. You jump straight to the solution of a problem/need that is worth introducing first, IMO.

timyates (approved):
Approve, but agree with @alvarosanchez's review

A paragraph explaining what we're trying to do, and then another at the end of the controller section explaining what we are going to do to fix it would improve it imho

https://github.com/micronaut-projects/micronaut-guides/blob/cc21564b1725f66451dec2c64e9038c9ef99635f/guides/micronaut-patterns-composite/micronaut-patterns-composite.adoc#L50-L74

sdelamo:
> The guide is great. But I miss an introduction paragraph. You jump straight to the solution of a problem/need that is worth introducing first, IMO.

@alvarosanchez  I have added an introduction paragraph

https://github.com/micronaut-projects/micronaut-guides/pull/902/commits/13f4789fd4b3932818bf7edd04b73f027b439fd0

Moreover, [I have removed tests from the asciidoc to reduce clutter](https://github.com/micronaut-projects/micronaut-guides/pull/902/commits/0ffd6bc9426364e4dc0e8776e3d5b038003ddc2b)

alvarosanchez (approved):

PR #919, opened 2022-05-05
Reinstate guides once 3.4.4 is out

timyates:
This is to check if the guides all work with 3.4.4-SNAPSHOT which includes this fix

https://github.com/micronaut-projects/micronaut-starter/pull/1208#event-6568443535


sdelamo (approved):

sdelamo:
💃🕺 @timyates

timyates:
I reset the version back from the 3.4.4-SNAPSHOT version, so this should be good to merge once we release 3.4.4 and update guides to use that

timyates:
It ignored the [skip-ci] in my commit, so I cancelled them for now

PR #910, opened 2022-04-28
Build Fixes

burtbeckwith:

PR #922, opened 2022-05-09
Reinstate temporarily disabled failing test

timyates:
We had an issue here with a timeout

https://github.com/micronaut-projects/micronaut-guides/pull/910#issuecomment-1113598043

Reinstating this test to see if it's still the case


timyates:
Ok, reinstated the test, and it passes...

Ran the app, and ran

```
curl -X POST \
    -H 'Content-Type: application/json' \
    -d '{"username":"tim", "password":"nope"}' \
    http://localhost:8080/login
```

And it returned 

```
{"message":"User Not Found","_links":{"self":{"href":"/login","templated":false}}}
```

As expected.

I believe `ldap://ldap.forumsys.com:389` must have been unavailable.

We *could* update the timeout, but if it was just hanging this wouldn't help, and would just pollute this working guide

burtbeckwith (approved):

PR #945, opened 2022-05-31
Openapi generator incorrect version for dependency fix

andriy-dmytruk:
Fix for https://github.com/micronaut-projects/micronaut-guides/issues/943

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -27,6 +27,10 @@ dependencies {
         exclude group: "org.codehaus.groovy"
     }
     implementation("org.openapitools:openapi-generator:$openapiGeneratorVersion")
+    // Explicitly specify the version for swagger-models to be used by the openapi generator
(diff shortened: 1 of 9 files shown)


buildSrc/build.gradle
```diff
@@ -27,6 +27,10 @@ dependencies {
         exclude group: "org.codehaus.groovy"
     }
     implementation("org.openapitools:openapi-generator:$openapiGeneratorVersion")
+    // Explicitly specify the version for swagger-models to be used by the openapi generator
+    implementation("io.swagger.core.v3:swagger-models") {
+        version { strictly "$swaggerModelsVersion" }
```
andriy-dmytruk, line 32: Setting the strict version solves the problem. It is not clear why gradle wants a different version for it (as it seems a much lower one, possibly even `io.swagger:swagger-models`). I couldn't find any other place where swagger models or core were added as dependency.

sdelamo (approved):

PR #981, opened 2022-06-17
landing page per tag, improve index

### LINKED ISSUE #661: Add tags
@graemerocher 

 > The PM of Kotlin pointed out that this link is broken https://guides.micronaut.io/tags/kotlin.html
> apparently it was linked from https://kotlinlang.org/docs/server-overview.html#deploying-kotlin-server-side-applications and used to work

sdelamo:
Close https://github.com/micronaut-projects/micronaut-guides/issues/661

### DIFF
--- a/buildSrc/src/main/groovy/io/micronaut/guides/IndexGenerator.groovy
+++ b/buildSrc/src/main/groovy/io/micronaut/guides/IndexGenerator.groovy
@@ -3,10 +3,14 @@ package io.micronaut.guides
 import groovy.json.JsonOutput
 import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
+import io.micronaut.core.order.OrderUtil
+import io.micronaut.core.order.Ordered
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/groovy/io/micronaut/guides/IndexGenerator.groovy
```diff
@@ -31,52 +36,99 @@ class IndexGenerator {
...
+            index += '    </div>'
+            index += '  </div>'
+            index += '</div>'
+            index += '</div>'
+            index += '<div class="categorygrid">'
+            index += '<div class="row">'
+            index += '    <div class="inner" style="padding: 0">'
+            index += TagCloud.tagCloud(tags)
+            index += '    </div>'
+            index += '</div>'
```
timyates: Think we need a new class on this, Row has

```
margin-right: -15px;
margin-left: -15px;
```

Which causes the tags to go outside of the DIV border

image

timyates (approved):

PR #997, opened 2022-07-01
Update HelloControllerSpec.groovy

mtalaga:
When I was trying to make rest-assured test work in Groovy/Gradle setup, any test was able to found through runner. After adding `expect:` clause code started to work.


sdelamo:
> When I was trying to make rest-assured test work in Groovy/Gradle setup, any test was able to found through runner. After adding `expect:` clause code started to work.

Thanks for the contribution!

sdelamo (approved):

PR #1003, opened 2022-07-06
Guide alternate styling

JasonTypesCodes:
Alternate styling options for the guide lists

image

image (1)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ buildSrc/src/generated
 *.scpt
 .nova
 .run
+
+cli/bin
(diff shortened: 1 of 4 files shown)


sdelamo (approved):

sdelamo:
Once we approve this PR, we will get rid of the guides index in wordpress.

timyates (approved):
On small screens, the blue title box goes a bit big?

image

If we want to reduce it's height, we can add something like:

```css
@media (max-width: 767.98px) {
    .grid-item:before {
        padding-top: 40%;
    }
}
```

To `guide.css` which would change it to

image

Or, we could leave it as is if that is preferred?

sdelamo:
please, apply the fix @timyates  

> ```css
> @media (max-width: 767.98px) {
>     .grid-item:before {
>         padding-top: 40%;
>     }
> }
> ```

timyates:
@sdelamo done 👍

PR #1068, opened 2022-09-08
Stop shared test resources server after running each test

timyates:
When we hit the messaging guides, we have them configured to share the server.

This is so that guides with multiple components that talk to each other work as expected (you don't want your Producer and Consumer looking at a different instance of Kafka)

This has the downside that we end up with lots of docker services hanging around, which I believe is introducing flakiness into the build.  Not sure whether this is caused by some sort of race condition, or (more likely) resource exhaustion when we have many services all running inside docker on the CI machine.

We also have many tests which expect a database or channel to be empty when they start, which is potentially not the case when running multiple guides using the same resource.

This change introduces an extra line to the test running script that attempts to shut down the shared server after each build

@melix will this work?  Or will it pull the plug on another test that may be running at the time?


timyates:
This isn't a total fix, as the TestResources server is shared among all projects on a machine.  This means we may incorrectly kill the server (if two messaging guides are running at the same time)

sdelamo:
I am going to merge this because this seems to helps with execution time. 

This is running for 5 hours: 

https://github.com/micronaut-projects/micronaut-guides/actions/runs/3020163153

PR #1072, opened 2022-09-09
build: Micronaut 3.6.3

sdelamo:


sdelamo:
@timyates can you investigate the failure?

timyates:
Failure passes locally... rerunning to see

timyates:
@sdelamo I think this is paralleism and Test Resources...

Should I turn off the parallel execution?

But without the "shut down resources" that was reverted, I'm not sure it will work...

timyates:
My suggestion is to:

1. Turn off parallelism in the RunTestScript task for now
1. Re-instate the "shut down test resources" but hide the output so people don't think it's failing the scripts

timyates:
🎉

PR #1080, opened 2022-09-13
Guide: How to use Micronaut Data from a Spring Boot app

sdelamo:
To generate the guide: 

```
./gradlew springBootMicronautDataBuild
```

To open the HTML: 

```
open build/dist/micronaut-data-jdbc-repository-gradle-java.html 
```
[Micronaut Data from a Spring Boot Application.pdf](https://github.com/micronaut-projects/micronaut-guides/files/9585500/Micronaut.Data.from.a.Spring.Boot.Application.pdf)


buildSrc/src/main/resources/pom.xml
```diff
@@ -12,6 +12,21 @@
         </repository>
     </repositories>
     <dependencies>
+        <dependency>
+            <groupId>jakarta.annotation</groupId>
+            <artifactId>jakarta.annotation-api</artifactId>
+            <version>2.1.1</version>
```
alvarosanchez, line 18: Do we really need those versions? They are in the BOM no?

alvarosanchez (approved):

buildSrc/src/main/resources/pom.xml
graemerocher: Spring Boot's BOM forces these to be downgraded without the explicit versions

alvarosanchez: Ok I see. In such case, wouldn't it be better to "enforce" our BOM if "micronaut-spring" is used?

sdelamo: Please note that we currently don't support generating `enforcedPlatform` dependencies in `micronaut-starter`

alvarosanchez: My point is that, if this is necessary for the guide example, isn't going to be necessary as well for starter apps?

Also, mixing Spring's and Micronaut's BOM can be a hide and seek game. What other dependencies might be being downgraded to an incompatible version because of that?

To be clear, I'm fine with this change as part of this guide. My only point is to be sure we aren't missing the same elsewhere.

graemerocher: I agree, but overriding all the versions in the Spring BOM with those in the Micronaut BOM may not be safe either. Either way this has been documented here as well https://micronaut-projects.github.io/micronaut-spring/latest/guide/#springBootStarter

graemerocher (approved):

buildSrc/src/main/resources/pom.xml
sdelamo: and the guide links to that part of the documentation.

PR #1093, opened 2022-09-18
improve AWS Object Storage guide test

sdelamo:
- Change delete endpoint to use @Status(NO_CONTENT)
- Use location header in the test
- Migrate test to JUnit
- Download file as byte[]
- adds LocalStack stop as a Junit Extensions


guides/micronaut-object-storage-aws/metadata.json
```diff
@@ -7,11 +7,10 @@
   "categories": ["AWS"],
   "publicationDate": "2022-10-01",
   "languages": ["java"],
-  "testFramework": "spock",
   "apps": [
     {
       "name": "default",
-      "features": ["object-storage-aws", "graalvm"]
+      "features": ["object-storage-aws", "graalvm", "localstack"]
```
alvarosanchez: I purposely didn't add the `localstack` feature explicitly because then it shows up in the guide:

image

alvarosanchez (changes requested):
There are few things that can be cleaned up. Also FYI, the test is not executed (`@Test` isn't inherited).

But don't worry, I'll take it over from here and fix it. Thank you very much!

guides/micronaut-object-storage-aws/metadata.json
sdelamo: I think I can fix that a separate issue. Not including local features in docs.

PR #1102, opened 2022-09-22
dont output local features to the userguide

sdelamo:


buildSrc/src/main/groovy/io/micronaut/guides/GuideAsciidocGenerator.groovy
```diff
@@ -626,4 +627,14 @@ class GuideAsciidocGenerator {
             return pomDependencyVersionResolver.getCoordinates()
         }
     }
+
+    private static List<String> guidesFeatures() {
+        try (ApplicationContext context = ApplicationContext.run()) {
```
alvarosanchez: Is there no other way? This seems overkilling.

What if `GuideMetadata` would be changed to have `localFeatures` in addition to `features`, so that only the latter would show up?

sdelamo: > Is there no other way? This seems overkilling.

Being a static it should run only once per build.

sdelamo: > What if GuideMetadata would be changed to have localFeatures in addition to features, so that only the latter would show up?

If I understand correct, @alvarosanchez means adding an extra entry in the metadata.json file. for example: `features` and `invisible-features`

sdelamo: Done via https://github.com/micronaut-projects/micronaut-guides/pull/1102/commits/c9e6acd6146c74065478a486e64d82b148348fc7

alvarosanchez (approved):

PR #1114, opened 2022-09-28
feat(auth0): use Authentication object in the example and display informations

adrienpessu:
As I am working on a Micronaut / Auth0 sample for the Auth0 website.
I came across this great guide. 

The `<span th:text="${security.attributes.get('email')}">` didn't work for me and I also wanted to show people how to get authentication informations.

Just let me know what you think.

### DIFF
--- a/guides/micronaut-oauth2-auth0/java/src/main/java/example/micronaut/HomeController.java
+++ b/guides/micronaut-oauth2-auth0/java/src/main/java/example/micronaut/HomeController.java
@@ -15,7 +15,13 @@ public class HomeController {
     @Secured(SecurityRule.IS_ANONYMOUS) // <2>
     @View("home") // <3>
     @Get // <4>
-    public Map<String, Object> index() {
-        return new HashMap<>();
+    public HttpResponse<Map<String, Object>> index(@Nullable Authentication authentication) {
+
+        HashMap<String, Object> variables = new HashMap<>();
+        if (authentication != null) {
+            variables.putAll(authentication.getAttributes());
+        }
+
+        return HttpResponse.ok(variables);
     }
 }
--- a/guides/micronaut-oauth2-auth0/java/src/main/java/example/micronaut/HomeController.java
+++ b/guides/micronaut-oauth2-auth0/java/src/main/java/example/micronaut/HomeController.java
@@ -17,7 +17,7 @@ public class HomeController {
     @Get // <4>
     public HttpResponse<Map<String, Object>> index(@Nullable Authentication authentication) {
 
-        HashMap<String, Object> variables = new HashMap<>();
+        Map<String, Object> variables = new HashMap<>();
         if (authentication != null) {
             variables.putAll(authentication.getAttributes());
         }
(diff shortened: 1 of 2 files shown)


guides/micronaut-oauth2-auth0/java/src/main/java/example/micronaut/HomeController.java
```diff
@@ -15,7 +15,13 @@ public class HomeController {
     @Secured(SecurityRule.IS_ANONYMOUS) // <2>
     @View("home") // <3>
     @Get // <4>
-    public Map<String, Object> index() {
-        return new HashMap<>();
+    public HttpResponse<Map<String, Object>> index(@Nullable Authentication authentication) {
+
+        HashMap<String, Object> variables = new HashMap<>();
```
burtbeckwith:
```suggestion
        Map<String, Object> variables = new HashMap<>();
```

guides/micronaut-oauth2-auth0/src/main/resources/views/home.html
```diff
@@ -6,14 +6,44 @@
 <body>
 <h1>Micronaut - Auth0 example</h1>
 
-<h2 th:if="${security}">username: <span th:text="${security.attributes.get('email')}"></span></h2>
-<h2 th:unless="${security}">username: Anonymous</h2>
+<h2 th:if="${security && nickname}">nickname: <span th:text="(${nickname})"></span></h2>
+<img th:if="${picture}" style="border-radius: 50%;" height="75" alt="Profil picture" th:src="${picture}"/>
```
burtbeckwith:
```suggestion
<img th:if="${picture}" style="border-radius: 50%;" height="75" alt="Profile picture" th:src="${picture}"/>
```

guides/micronaut-oauth2-auth0/java/src/main/java/example/micronaut/HomeController.java
```diff
@@ -15,7 +15,13 @@ public class HomeController {
     @Secured(SecurityRule.IS_ANONYMOUS) // <2>
     @View("home") // <3>
     @Get // <4>
-    public Map<String, Object> index() {
-        return new HashMap<>();
+    public HttpResponse<Map<String, Object>> index(@Nullable Authentication authentication) {
```
sdelamo, line 18: This change should not be necessary. 

We have a View Model Processor which includes the security in the model map. 

https://github.com/micronaut-projects/micronaut-views/blob/master/views-core/src/main/java/io/micronaut/views/model/security/SecurityViewModelProcessor.java

I am surprised it did not work for you. 

I will be happy to show more information in the home.html screen but I would like not do changes to the controller if possible.

sdelamo (changes requested):

guides/micronaut-oauth2-auth0/java/src/main/java/example/micronaut/HomeController.java
adrienpessu: Ok, now, I know why the parameter wasn't reachable in the `Thymeleaf` template. I was returning en empty response `return HttpResponse.ok();`. In that case, `${security.parameters}` is empty.
Nevertheless, I founded useful to know that we can access information in the controller and in the template. But if you don't, I understand. 

Thank you for the link to `SecurityViewModelProcessor`

PR #1123, opened 2022-10-17
Switch to Serde for MicroStream guide

timyates:
~~I had to disable Groovy due to https://github.com/micronaut-projects/micronaut-core/issues/8184~~

I re-enabled Groovy, and added problem-json to work around the above issue


guides/micronaut-microstream-persistence/java/src/main/java/example/micronaut/FruitContainer.java
```diff
@@ -1,12 +1,12 @@
 package example.micronaut;
 
-import io.micronaut.core.annotation.Introspected;
 import io.micronaut.core.annotation.NonNull;
+
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
-@Introspected // <1>
```
timyates, line 8: The Root Object does not need to be introspected or serdeable

timyates:
I *think* the mongo sync guide is failing as the connection to docker is too slow 🤔

sdelamo (approved):

PR #1140, opened 2022-10-21
Switch to use Serde for the Data Hibernate Reactive guide

timyates:


guides/micronaut-data-hibernate-reactive/java/src/main/java/example/micronaut/domain/Genre.java
```diff
@@ -8,6 +9,7 @@
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 
+@Serdeable
```
timyates, line 12: I cannot see why we now need to add `@Serdeable` here, but previously it worked without `@Introspected`

sdelamo (approved):

PR #1146, opened 2022-10-26
Switch to use Serde for the Hibernate Reactive JPA guide

timyates:


timyates:
Doesn't currently work.

The guide didn't use Test Resources, so I have moved to use this as well as Serde.

I need to find out how to configure TR for Hibernate Reactive and JPA
Currently disabled Kotlin whilst I confer on this

timyates:
Backed out the Test Resources change, so this just switched to use Serde, fixes the callouts in the guide, and I formatted the Kotlin better (IMHO)

sdelamo (approved):

PR #1164, opened 2022-11-17
Use the correct version for the micronaut-data processor with SpringBoot

### LINKED ISSUE #1163: Make test action stable
https://github.com/micronaut-projects/micronaut-guides/actions/workflows/snapshot.yml

if we need to run only 17, I am fine with it.

timyates:
We were using the version of Micronaut, which is incorrect


sdelamo (approved):

sdelamo:
> We were using the version of Micronaut, which is incorrect

Good catch. I guess sometimes was passing because it was aligned.

PR #1228, opened 2023-01-17
add lateinit and @field callouts

sdelamo:


graemerocher:
fine to call this out although it is probably not best practise to use field injection. You can inject the client into the method or the constructor equally

graemerocher (approved):

PR #1236, opened 2023-02-08
Mock Github Server to avoid rate limiting in HTTP Client Guide

sdelamo:
This PR updates to 3.8.4


guides/micronaut-http-client/groovy/src/test/groovy/example/micronaut/GithubControllerSpec.groovy
```diff
@@ -1,60 +1,77 @@
...
+        }
 
-        then:
-        githubReleases.stream()
-                .map(GithubRelease::getName)
-                .allMatch(name -> MICRONAUT_RELEASE.matcher(name)
-                        .find())
+        @Consumes(value = ["application/vnd.github.v3+json", "application/json"])
+        @Get("/repos/micronaut-projects/micronaut-core/releases") // <3>
+        String coreReleases() {
```
sdelamo, line 71: I don't know if it is Apache Groovy or a bug in Micronaut Framework. If I set the same as in Kotlin or Java (see below), it does not work. 

```groovy
Optional<InputStream> coreReleases() {
            resourceLoader.getResourceAsStream("releases.json")
}

melix: I've seen cases where Groovy didn't resolve resources with the same path. You could try to use an absolute path to the resource instead.

melix (approved):

guides/micronaut-http-client/groovy/src/test/groovy/example/micronaut/GithubControllerSpec.groovy
sdelamo: Please, note that `resourceLoader.getResourceAsStream("releases.json")` seems to work. In order to make it work, in Apache Groovy I have to unwrap the optional and get the string of input stream with Groovy's `.text`. Thus, I think it is not an issue with the path. It seems the issue is returning an `Optional<InputStream>` with Apache Groovy.

timyates (approved):

guides/micronaut-http-client/groovy/src/test/groovy/example/micronaut/GithubControllerSpec.groovy
melix: Ah sorry I missed it wasn't a `ClassLoader` here but a resource loader.

PR #1237, opened 2023-02-08
X-API-KEY guide

sdelamo:
[Micronaut Security API Key.pdf](https://github.com/micronaut-projects/micronaut-guides/files/10686259/Micronaut.Security.API.Key.pdf)

### DIFF
--- a/src/docs/common/callouts/callout-each-property.adoc
+++ b/src/docs/common/callouts/callout-each-property.adoc
@@ -1,0 +1,1 @@
+The `@EachProperty` annotation creates a `ConfigurationProperties` bean for each sub-property within the given property.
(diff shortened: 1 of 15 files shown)


src/docs/common/callouts/callout-each-property.adoc
```diff
@@ -0,0 +1 @@
+The `@EachProperty` annotation creates a `ConfigurationProperties` bean for each sub-property within the given property.
```
alvarosanchez:
```suggestion
The `@EachProperty` annotation creates a `ConfigurationProperties` bean for each sub-property within the given name.
```

src/docs/common/callouts/callout-principal.adoc
```diff
@@ -0,0 +1 @@
+You can bind java.security.Principal as a method’s parameter in a controller.
```
alvarosanchez:
```suggestion
You can bind `java.security.Principal` as a method’s parameter in a controller.
```

alvarosanchez (approved):

src/docs/common/callouts/callout-each-property.adoc
sdelamo: @alvarosanchez I think we should change it in core as well https://docs.micronaut.io/latest/guide/#eachProperty. Thoughts?

alvarosanchez: Fine to me. It's just to not have the word "property" 3 times.

PR #1243, opened 2023-02-14
create the http client lazyly

sdelamo:


timyates (approved):
Is it simpler if we inject `EmbeddedServer` instead of `BeanContext`?

Then we don't need to get the EmbeddedServer bean

PR #1256, opened 2023-03-15
move create your first graalvm guide to distribution category

sdelamo:


graemerocher:
why?

sdelamo:
> why?

Because, we have a distribution category (FAT JAR, docker Image) and I think it fits better there so that users discover they can distribute as a native executable.

graemerocher (approved):

PR #1275, opened 2023-05-05
Add a section to the CORS guide about CrossOrigin annotation.

### LINKED ISSUE #1271: Add a section to the CORS guide about new annotation
https://guides.micronaut.io/latest/micronaut-cors.html

wetted:
closes #1271

### DIFF
--- a/guides/micronaut-cors/groovy/index.html
+++ b/guides/micronaut-cors/groovy/index.html
@@ -10,7 +10,7 @@
 			};
 			xhr.send();					
 		}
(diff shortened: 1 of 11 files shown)


wetted:
@sdelamo I wasn't sure how I could include the `hello-base` guide but without the controller, so I had to make changes that seemed unnecessarily complicated, such that the RunTestScript would pass, and the `@CrossOrigin` annotation would be included in the solution at the same time. So maybe this is a little too fussy.

PR #1284, opened 2023-05-24
Manually went through the K8S guide

timyates:
Awesome work, really good guide

But found a few snags and issues, which are fixed here

### DIFF
--- a/guides/micronaut-k8s/api/src/main/resources/application.yml
+++ b/guides/micronaut-k8s/api/src/main/resources/application.yml
@@ -1,3 +1,6 @@
+micronaut:
+  application:
+    name: api
 authentication-credentials:
   username: ${username} # <1>
   password: ${password} # <2>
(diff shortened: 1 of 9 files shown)


guides/micronaut-k8s/api/src/main/resources/application.yml
```diff
@@ -1,3 +1,6 @@
+micronaut:
+  application:
+    name: api
```
sdelamo, line 3: this should not be necessary @timyates since it is defined in bootstrap.yml

guides/micronaut-k8s/orders/src/main/resources/application.yml
```diff
@@ -1,3 +1,6 @@
+micronaut:
+  application:
+    name: orders
```
sdelamo, line 3: this should not be necessary, it is defined in bootstrap.yml

sdelamo (changes requested):

n0tl3ss (approved):

guides/micronaut-k8s/api/src/main/resources/application.yml
timyates: Yes, but the applications generate different contents (with this) when you create them

It's not necessary but it's a lot more obvious in the guide as to what you have to do

PR #1297, opened 2023-07-05
Re-introduce OpenAPI Client guide

melix:
This commit introduces a new guide for generation of OpenAPI clients. Compared to the previous version, this guide is much simplified:

- it doesn't make use of the Twitter API anymore, but uses OpenMeteo instead
- it doesn't require authentication, which was making the guide more complicated than it should
- it doesn't explain how to use the client in a controller, since this is generic Micronaut code and that we show how to use the client in tests


andriy-dmytruk (approved):

sdelamo (approved):
I have:

- added a missing configuration section. 
- removed micronaut.server.port=8080  That it is the default
- done slight changes to copy

PR #1298, opened 2023-07-07
Reintroduce OpenAPI server generation guide

melix:
This is now using the build plugins instead of the external OpenAPI generator project.

Closes https://github.com/micronaut-projects/micronaut-openapi/issues/1037


guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,550 @@
...
+<1> The api package will contain generated interfaces for our API that we will need to implement in our controller
+<2> It also generates model classes which are required by the service
+
+:exclude-for-build:
+
+:exclude-for-build: gradle
+
+Enable the Micronaut OpenAPI server generation in your `pom.xml` file, configure the location to the specification file and package names:
+
+[source, xml]
```
timyates, line 205: This ends up in the Gradle guide still

timyates: Fixed in https://github.com/micronaut-projects/micronaut-guides/pull/1298/commits/7ec014397a5faf47a675e64a4cbcfef6c1446915

timyates (approved):

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/BookEntity.java
```diff
@@ -0,0 +1,20 @@
...
+import io.micronaut.data.annotation.Id;
+import io.micronaut.data.annotation.MappedEntity;
+import io.micronaut.data.annotation.MappedProperty;
+
+import jakarta.validation.constraints.NotNull;
+
+@MappedEntity("book") // <1>
+public record BookEntity(
+    @Nullable @Id @GeneratedValue(GeneratedValue.Type.AUTO) Long id, // <2>
+    @NonNull @NotNull String name,
```
sdelamo:
```suggestion
    @NonNull @NotBlank String name,
```

Please, use @NotBlank instead of `@NotNull` for strings.

sdelamo: Use `@NotBlank` not `@NotNull`

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/controller/BooksController.java
```diff
@@ -0,0 +1,86 @@
...
+    //tag::inject[]
+    private final BookRepository bookRepository; // <1>
+
+    public BooksController(BookRepository bookRepository) { // <1>
+        this.bookRepository = bookRepository;
+    }
+    //end::inject[]
+
+    //tag::addBook[]
+    public void addBook(BookInfo bookInfo) {
```
sdelamo, line 44: misses `@ExecuteOn`

guides/micronaut-openapi-generator-server/java/src/main/resources/application.properties
```diff
@@ -0,0 +1,18 @@
+context-path=/
+micronaut.application.name=openapi-micronaut
+micronaut.server.port=8080
+#jackson.serialization.writeEnumsUsingToString=true
```
sdelamo: why add `micronaut.server.port=8080` That it is the default

guides/micronaut-openapi-generator-server/java/src/main/resources/application.properties
```diff
@@ -0,0 +1,18 @@
+context-path=/
```
sdelamo, line 2: why add `context.path`?

guides/micronaut-openapi-generator-server/java/src/main/resources/application.properties
```diff
@@ -0,0 +1,18 @@
...
+micronaut.application.name=openapi-micronaut
+micronaut.server.port=8080
+#jackson.serialization.writeEnumsUsingToString=true
+#jackson.serialization.writeDatesAsTimestamps=true
```
sdelamo: why add all of these entries as comments?

guides/micronaut-openapi-generator-server/java/src/test/java/example/micronaut/controller/BooksControllerTest.java
```diff
@@ -0,0 +1,45 @@
...
+import jakarta.inject.Inject;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@MicronautTest // <1>
+public class BooksControllerTest {
+
+    @Inject
+    @Client("${context-path}")
```
sdelamo, line 21: why `the context-path` ?

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/BookEntity.java
melix: Fixed in 299fb2e5a6

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,550 @@
...
+[source, groovy]
+.build.gradle
+----
+micronaut {
+  ...
+  openapi {
+    server(file("src/main/resources/library-definition.yml")) { // <1>
+      apiPackageName = "example.micronaut.api"
+      modelPackageName = "example.micronaut.model"
+      invokerPackage = "example.micronaut.invoker"
```
andriy-dmytruk: The invoker package is different for gradle and maven. Probably best to set it to simply `example.micronaut` common package.

The structure will be:
```
----
build/generated/openapi
├── generateServerOpenApiApis
│   └── src
│       └── main
│           └── java
│               └── example
│                   └── micronaut
│                       ├── annotation
│                       │   └── HardNullable.java
│                       └── api
│                             └── BooksApi.java        # <1>
└──  generateServerOpenApiModels
    └── src
        └── main
            └── java
                └── example
                    └── micronaut
                        └── model                              # <2>
                            ├── BookAvailability.java
                            └── BookInfo.java
----
```
and
```
----
target/generated-sources/openapi
└── src
    └── main
        └── java
            └── example
                └── micronaut
                    ├── annotation
                    │   └── HardNullable.java
                    ├── api                             # <1>
                    │   └── BooksApi.java
                    └── model                           # <2>
                        ├── BookAvailability.java
                        └── BookInfo.java
----
```

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
```diff
@@ -0,0 +1,550 @@
...
+that satisfies your needs. If you want to finish implementing the basic authentication, continue to the
+link:https://guides.micronaut.io/latest/micronaut-security-basicauth.html[Micronaut Basic Auth guide] and replicate
+steps to create the `AuthenticationProvider` and appropriate tests.
+
+NOTE: You can also read link:https://micronaut-projects.github.io/micronaut-security/latest/guide/[Micronaut Security documentation]
+or link:https://micronaut.io/guides[Micronaut guides] about security to learn more about
+the supported Authorization strategies.
+
+=== Generate Micronaut Client
+
```
andriy-dmytruk, line 553: We could add this info:
```
To specify the server URL to be consumed by OpenAPI client generators, use the link:servers list property[https://swagger.io/specification/#server-object] of the OpenAPI specification.
```

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
melix: I'm confused, where does the `annotation` package come from?

andriy-dmytruk: It comes from inside the invoker package. The invoker package has all the invocation and utility classes (had the `Application.java` previously and can have some security configuration for the client).
The `HardNullable` is an inherited nullable generated in `{invokerPackage}.annotation`.

melix: I understand but you said "use `micronaut.example` for the example package", not `micronaut.example.annotation`.

andriy-dmytruk: The `HardNullable` is put in a separate subpackage to make the project better structured in case there are other files in the invoker package (in this case there are none, so it looks strange).

melix: Oh. That's quite surprising behavior. But I think it won't work here, since we're generating classes in an isolated directory in `build/...` so there won't be other types in that package.

andriy-dmytruk: Currently, it works fine:
image

As I understand this PR is not used yet: https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/758. But it seems like [this line](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/204702be5d866419e3b577bb8990a9b41e0cd634/openapi-plugin/src/main/java/io/micronaut/gradle/openapi/DefaultOpenApiExtension.java#L80) adds supporting files to the api output, so the `HardNullable` should be generated then.

melix: the PR is used. I am extremely confused by the behavior. You are saying, basically, that if the invoker package is set to `example.micronaut`, then the annotation should be generated under `example.micronaut.annotation` _because_ there are other files in that directory. That is quite surprising behavior to me, as I would expect the generator to honor what I tell it to use. I also understand that "not empty" only works because by accident, the `api` and `model` types are created before the supporting files. This wouldn't be the case, for example, if we had 2 tasks, one to generate the `api`, the other to generate the supporting files. So I would rather use a `example.micronaut.invoker` package, so that the behavior is consistent.

andriy-dmytruk: No, it is just generated in the `example.micronaut.annotation` directory solely based on the fact that the invoker package is `example.micronaut` and independently of if there are other files in the directory.

melix: Then I don't understand. Why doesn't it generate something in `example.micronaut.invoker.annotation` if I set the package to `example.micronaut.invoker`? This seems to contradict what I'm effectively seeing.

andriy-dmytruk: Sorry for the confusion... This is what it will generate if we set the invoker package to `example.micronaut`:
> The invoker package is different for gradle and maven. Probably best to set it to simply example.micronaut common package.

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
melix: mmm, not sure it's useful information for the server guide, is it? We could do this in the client guide, but I think we already support setting the URL. BTW I think we should make the Micronaut property that sets the URL configurable in the generator:

```
@Client("${openapi-micronaut-client-base-path}")
```

because you may want to have multiple clients for different services in a single project, in which case you want to use different properties.

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
melix: Ok thanks, I fixed the (non intended) difference in the Maven example, and used `example.micronaut` in both cases for the invoker package.

andriy-dmytruk: Thank you!

guides/micronaut-openapi-generator-server/micronaut-openapi-generator-server.adoc
andriy-dmytruk: Currently it can be changed by configuring the `artifactId` property. But I agree that we should improve this. We should use micronaut configuration via client id instead: https://docs.micronaut.io/latest/guide/#_client_specific_configuration (`micronaut.http.services.client-id.urls` for `@Client("[client-id]")`.

andriy-dmytruk (approved):

guides/micronaut-openapi-generator-server/java/src/main/resources/application.properties
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/1298/commits/9ff2a1d3c7c9e10e4cbdbee6f27f067d48b63490

sdelamo (approved):

guides/micronaut-openapi-generator-server/java/src/main/resources/application.properties
sdelamo: removed them here: https://github.com/micronaut-projects/micronaut-guides/pull/1298/commits/ed6f0b6efee07697a6b41f160dff8e464ebea0b6

guides/micronaut-openapi-generator-server/java/src/main/java/example/micronaut/controller/BooksController.java
sdelamo: resolved via https://github.com/micronaut-projects/micronaut-guides/pull/1298/commits/6bcb8b05478fb66fcf782a6d590b3ecbceb89e6d

PR #1302, opened 2023-07-22
Consuming a REST API - SpringBoot vs Micronaut Framework

sdelamo:


guides/spring-boot-to-micronaut-consuming-rest/springboot/java/src/main/java/example/micronaut/Application.java
```diff
@@ -0,0 +1,34 @@
...
+    public RestTemplate restTemplate(RestTemplateBuilder builder) {
+        return builder.build();
+    }
+
+    @Bean
+    public CommandLineRunner run(RestTemplate restTemplate) throws Exception {
+        return args -> {
+            Quote quote = restTemplate.getForObject(
+                    "http://localhost:8080/api/random", Quote.class);
+            log.info(quote.toString());
```
graemerocher, line 31: in the latest version of spring they have a  declarative http client too which would be a more accurate comparison

guides/spring-boot-to-micronaut-consuming-rest/micronautframework/java/src/main/java/example/micronaut/Application.java
```diff
@@ -0,0 +1,30 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(Application.class);
+    public static void main(String[] args) {
+        Micronaut.run(Application.class, args);
+    }
+
+    @Inject // <2>
+    QuotersClient quotersClient;
+
+    @Override
+    public void onApplicationEvent(StartupEvent event) {  // <1>
```
graemerocher, line 21: it is more idiomatic to use `@EventListener`

PR #1319, opened 2023-08-22
Update guides that explicitly show --jdk=11

timyates:


sdelamo (approved):
Maybe we should just removed that.

PR #1331, opened 2023-09-14
update TeamConfigurationTest with more idiomatic Kotlin

codewithbear:
Cleaning up TeamConfigurationTest to use Kotlin code instead of java classes
- use listOf and mapOf instead of java classes
- simplify forEach loop
- remove unnecessary imports for java classes


sdelamo (approved):

sdelamo:
@codewithbear thanks for the contribution!

PR #1334, opened 2023-09-15
Working with jOOQ Flyway using Testcontainers

sdelamo:
Port of Spring Boot Guide https://testcontainers.com/guides/working-with-jooq-flyway-using-testcontainers

I added a Test Resources section to show how you can simplify testing.

To generate guide locally:

./gradlew workingWithJooqFlywayUsingTestcontainersBuild

Open HTML locally:

open build/dist/index.html

Code available:

build/code/working-with-jooq-flyway-using-testcontainers


sdelamo:
@sivaprasadreddy  can you review this?

timyates:
We _could_ get Gradle working, but it would need a change in starter

If we generate a build with [this plugin](https://github.com/monosoul/jooq-gradle-plugin)

```
    id("dev.monosoul.jooq-docker") version "5.0.7"
```

Some config:

```
configurations {
    jooqCodegen {
        extendsFrom(implementation)
    }
}

tasks {
    generateJooqClasses {
        basePackageName.set("example.micronaut.jooq")
    }
}
```

And an additional dependency:

```
    jooqCodegen("org.postgresql:postgresql")
```

Then the Gradle build works...  However I can't find a way to add the config or dependency without changing starter 🤔

timyates (approved):

sivaprasadreddy (approved):
LGTM

melix (approved):
Something is unclear to me. What do we recommend? Using test containers directly or test resources? If we want to use test resources, why explain how to do it with test containers before?

sdelamo:
> Something is unclear to me. What do we recommend? Using test containers directly or test resources? If we want to use test resources, why explain how to do it with test containers before?

We recommend test resources. Explaining how to do it with test containers first shows what Test Resources gives you out-of-the-box (no testcontainers dependencies, more straightforward tests to write, etc.).

melix:
Then I would clarify this in the intro section:

```
== What we are going to achieve in this guide
We are going to create a Micronaut project using https://www.jooq.org/[jOOQ] together with https://www.postgresql.org/[PostgreSQL].
We will create our database tables by using https://flywaydb.org/[Flyway] Database migrations.
We will configure https://github.com/testcontainers/testcontainers-jooq-codegen-maven-plugin[testcontainers-jooq-codegen-maven-plugin] to generate jOOQ code using Testcontainers and Flyway
migration scripts.
```

This doesn't mention test resources, which is the final state of the guide. I think we should mention that we're going to simplify somehow.

PR #1336, opened 2023-09-18
Small improvements in Http Client Guide

codewithbear:
A couple small improvements in Http Client Guide

- Fix regular expression in GithubControllerTest. It works against the existing releases.json in the repo, however it fails against the latest releases from Github.
- Simplify reading `releases.json` in Kotlin version of the test, so that it's done with less code. Also throwing instead of returning Optional. In case of missing releases.json the test would fail either way, but with `orElseThrow` it highlights the line with releases.json, easier to troubleshoot.
- Mention in the guide the user needs to create releases.json themselves. Now it might not be clear right away


sdelamo (approved):
♥️ the improvements.

PR #1337, opened 2023-09-18
fix cli command for guides with junit-params

codewithbear:
Trying to execute this guide and getting an error
https://guides.micronaut.io/latest/micronaut-custom-validation-annotation-gradle-java.html

```
| Error junit-params requires JUnit.
```

Looks like specifying test framework is mandatory when adding junit-params as feature. It makes the cli command for generating the project invalid in a couple of places.

This PR adds a template for cli when the test framework needs to be specified explicitly.

How it looks like after the change:
image


sdelamo (changes requested):
Lets add 

```
    --test=@testFramework@
```

to `common-create-app-features.adoc` instead of creating a new file.

codewithbear:
@sdelamo sounds good, fixed now

sdelamo (approved):

PR #1343, opened 2023-09-21
simplify java code in validation guide

codewithbear:
Using java streams to reduce the amount of boilerplate code in the validation guide


sdelamo (approved):
Thanks!

PR #1352, opened 2023-09-27
For Micronaut 4.1.4 re-enable maven tests

timyates:
That should be fixed by 4.1.3

Added to the 4.1.3 project for now, but should be in 4.1.4


sdelamo:
I just commit 4.1.3 in this branch

sdelamo (approved):

timyates:
The `creating-your-first-micronaut-app-gradle-groovy` guide passes if we add `compileOnly("io.micronaut:micronaut-http-client")`

This will be fixed by https://github.com/micronaut-projects/micronaut-starter/pull/2096

sdelamo:
I have added a replacement.

timyates:
@sdelamo there's been a change in core so we need to mark the `@QueryValue` as `@Nullable` now in the `micronaut-graalvm-reflection` guide

https://ge.micronaut.io/s/rbxsvu2rausvc/console-log?anchor=4212&page=5

timyates:
That guide shouldn't be on this branch...  🤔. Re-running...

timyates:
It's because it's testing a merge 🤷‍♀️

Merged master in to here, and added a fix https://github.com/micronaut-projects/micronaut-guides/pull/1352/commits/b278fdf11f2d07059a40ce0f8b7d54cd6615a1fa

I am not sure why this fix is required...

timyates:
So we have a test here which tests exactly what we're seeing here

https://github.com/micronaut-projects/micronaut-core/blob/d08b5d49f76283d36a2654ef04c5ddcaf1cd2764/http-validation/src/test/groovy/io/micronaut/validation/routes/NullableParameterRuleSpec.groovy#L111

That without `@Nullable` we get an exception

PR #1353, opened 2023-09-28
GraalVM Polyglot JS

sdelamo:
To generate guide locally:

`./gradlew micronautGraalvmPolyglotJsBuild`
Open HTML locally:

`open build/dist/index.html`

Code available:

`build/code/micronaut-graalvm-polyglot-js`

Related to: 

- [Truffle unchained Blog post](https://medium.com/graalvm/truffle-unchained-13887b77b62c)
• [Example Repository](https://github.com/graalvm/polyglot-embedding-demo)
• [Embedding Guide](https://www.graalvm.org/latest/reference-manual/embed-languages/)
• [23.1 Release Announcement](https://medium.com/graalvm/new-truffle-and-graalvm-languages-release-1e5c8cdaaff)
• [Module Migration for Language Implementations](https://github.com/oracle/graal/blob/master/truffle/docs/ModuleMigration.md)

### DIFF
--- a/buildSrc/src/main/resources/pom.xml
+++ b/buildSrc/src/main/resources/pom.xml
@@ -13,6 +13,16 @@
     </repositories>
     <dependencies>
         <dependency>
+            <groupId>org.graalvm.polyglot</groupId>
+            <artifactId>polyglot</artifactId>
+            <version>23.1.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.graalvm.polyglot</groupId>
+            <artifactId>js</artifactId>
+            <version>23.1.0</version>
+        </dependency>
+        <dependency>
             <groupId>org.mock-server</groupId>
             <artifactId>mockserver-client-java</artifactId>
             <version>5.15.0</version>
--- a/buildSrc/src/main/java/io/micronaut/guides/feature/GraalvmPolyglot.java
+++ b/buildSrc/src/main/java/io/micronaut/guides/feature/GraalvmPolyglot.java
@@ -1,0 +1,10 @@
+package io.micronaut.guides.feature;
+
+import jakarta.inject.Singleton;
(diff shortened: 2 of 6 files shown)


sdelamo:
@chumer lets revisit this sample guide once we have support from [Java 21 in Micronaut framework](https://github.com/micronaut-projects/micronaut-core/issues/9869).

buildSrc/src/main/resources/pom.xml
```diff
@@ -12,6 +12,16 @@
...
     <dependencies>
+        <dependency>
+            <groupId>org.graalvm.polyglot</groupId>
+            <artifactId>polyglot</artifactId>
+            <version>23.1.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.graalvm.polyglot</groupId>
+            <artifactId>js</artifactId>
+            <version>23.1.0</version>
```
chumer, line 23: This needs a 
```xml
<type>pom</type>
```

chumer:
@sdelamo I've opened https://github.com/micronaut-projects/micronaut-maven-plugin/issues/866 to start the discussion. For the sample project attached in the issue Micronaut already works with GraalVM for JDK 21, so I think we already can start testing.

buildSrc/src/main/resources/pom.xml
sdelamo: we don't use this pom.xml as a typical pom.xml but as a way define versions coordinates.

chumer: Sorry, I cannot follow. Afaik with Maven, you need to specify that a dependency is of type pom if the target dependency is a pom type dependency on Maven central (which is the case for org.graalvm.polyglot:js). Otherwise it fails badly.

PR #1362, opened 2023-10-12
Run tests with Java 21

### LINKED ISSUE #1329: Test guides with Java 21
I am not sure we should test with both 17 and 21 in every PR. But we could have a cron Github Action which tests daily with 21. Thoughts?

timyates:
Once starter is released for 4.2.0 and we can generate Java 21 projects, this can be merged

Currently it is set up to use 4.2.0-SNAPSHOT as the version target.

Excluded lambda and function guides which only work on 17.

There is one odd change that was required.  When running Kapt projects, we needed to kill the Kotlin Compile Daemon as it holds on to the old properties it was started with, and doesn't notice that it needs to update to include the `--add-opens` for Kapt to work.

- We can't `--stop` the gradle daemons (as this would kill the build itself)
- We cannot use `--no-daemon` for all the builds as it takes AGES to run all the tests
- We cannot use `--no-daemon` on just the kapt builds, as the Gradle daemon finds the old Kotlin daemon.

So I went with `find the kotlin daemon and kill it` before and after Kapt builds


buildSrc/src/main/java/io/micronaut/guides/feature/MockServerClient.java
```diff
@@ -1,44 +0,0 @@
...
-import io.micronaut.core.annotation.Nullable;
-import io.micronaut.starter.application.ApplicationType;
-import io.micronaut.starter.application.generator.GeneratorContext;
-import io.micronaut.starter.build.dependencies.Dependency;
-import io.micronaut.starter.feature.Feature;
-import io.micronaut.starter.feature.database.TestContainers;
-import jakarta.inject.Singleton;
-
-@Singleton
-public class MockServerClient implements Feature {
```
sdelamo: was this not used?

buildSrc/build.gradle
```diff
@@ -10,6 +10,7 @@ ext {
 }
 
 repositories {
+    mavenLocal()
```
sdelamo: remove `mavenLocal()`

build.gradle
```diff
@@ -13,6 +13,7 @@ apply from: "gradle/asciidoc.gradle"
 version '0.0.1'
 
 repositories {
+    mavenLocal()
```
sdelamo: remove `mavenLocal()`

sdelamo (changes requested):

buildSrc/build.gradle
timyates: Already gone :-)

buildSrc/src/main/java/io/micronaut/guides/feature/MockServerClient.java
timyates: It collides with the one in v4.2.0 of Micronaut Starter

https://github.com/micronaut-projects/micronaut-starter/blob/4.2.x/starter-core/src/main/java/io/micronaut/starter/feature/test/MockServerClient.java

I assume this was added as a stop-gap until we got to use the one in 4.2.0?

sdelamo (approved):

timyates:
@sdelamo should we switch to just testing 21 on a nightly basis?

sdelamo:
> @sdelamo should we switch to just testing 21 on a nightly basis?

yup. 

great job making the java 21 transition

PR #1367, opened 2023-10-26
use properties configuration in okta guide

sdelamo:


timyates (approved):
I think the callouts were wrong.  I [pushed a fix](https://github.com/micronaut-projects/micronaut-guides/pull/1367/commits/2fb40c593f9538e31a88c935f7f576887f7b91e6), please check it @sdelamo 👍

sdelamo:
> I think the callouts were wrong. I [pushed a fix](https://github.com/micronaut-projects/micronaut-guides/pull/1367/commits/2fb40c593f9538e31a88c935f7f576887f7b91e6), please check it @sdelamo 👍

I think they were better previously. it is a problem with properties files where we cannot have inline comments. Can you revert your commit.

timyates:
You mean this is right? see item 2?

image

sdelamo:
> You mean this is right? see item 2?
> 
> image

I want to convey the user that `okta` in  `micronaut.security.oauth2.clients.okta.client-secret` is key.

timyates:
Ok, pushed the revert.   was confused as `The provider identifier should match the last part of the URL you entered as a redirect URL "/oauth/callback/okta"` didn't make sense to me, and then item 3 talks about the secret which is shown in item 2 in the source

PR #1372, opened 2023-11-17
run jooq tests

sdelamo:


timyates:
So this is failing as we removed `testcontainers.version` from the platform 😢 

@alvarosanchez @melix Obviously, we don't want this hardcoded here, but I can't find the magic Maven xml to look this up from https://search.maven.org/artifact/io.micronaut.testresources/micronaut-test-resources-bom/2.2.0/pom

Any ideas?

timyates:
Hmmm... looks like properties are only inherited from the parent...? 😭

alvarosanchez:
There's no possible magic, it's not possible. Properties are only inherited from the parent.

timyates:
Maybe hardcoding is the way forward for this case... 🤔

timyates:
@melix would it make sense to pull all module bom properties into the `io.micronaut.platform:micronaut-parent` (or Álvaro says maybe `micronaut-platform`) pom?

Or at leaset testcontainers.version? 😉

melix:
I am not sure it's a good practice, I will let @alvarosanchez decide.

alvarosanchez:
The thing is that 4.x made a breaking change, since previously all those properties were available in the old uber BOM. Inlining all properties in `micronaut-platform` BOM (not the parent POM) will restore this behaviour.

However, now that I think about it, perhaps it wouldn't work since the individual module BOMs are now _imported_ and not _inherited from_.

@timyates did you try setting the `<testcontainers.version>` property?

timyates:
> @timyates did you try setting the `<testcontainers.version>` property?

To this repo [like this?](https://github.com/micronaut-projects/micronaut-guides/pull/1372/files#diff-37af07b0c4579109ac61bae1d6944d5cdec2a97f47205a314a7b807e1429171fR34)

alvarosanchez:
Yes, to see if the property defined in the generated application is used when importing the BOM.

timyates:
It passed locally...  Of course, it's a buildSrc change so we'll have to wait a bit for all the tests to run again to see it pass on CI ;-)

[edit] Aaaand it's failing on CI :-(

Think it ran out of disk space 🤔 rerunning

timyates:
✅

PR #1379, opened 2023-11-22
Micronaut CRaC JDBC Guide

sdelamo:


timyates:
The maven guide seems to hang for me... investigating

timyates:
I think there's something odd going on in the maven test...  Sometimes it passes, sometimes it fails, and sometimes it hangs...

And I cannot think why

sdelamo:
> I think there's something odd going on in the maven test... Sometimes it passes, sometimes it fails, and sometimes it hangs...
> 
> And I cannot think why

should we remove maven for now?

timyates:
I'm not sure if it's just maven... 🤔

timyates (changes requested):
Something is wrong here.  Both Maven and Gradle are sometimes hanging when the tests are executed multiple times...

As far as I can tell, it gets down to calling `server.start()` in the CRaC `NettyEmbeddedServerResource`, but then it never comes back.

I tried with both TestResources and TestContainers.  Something must have changed in netty that's causing the hanging...

PR #1380, opened 2023-11-22
eclipestore guide

sdelamo:


timyates (approved):
Do we need to update the ExecuteOn in the MicroStream guide to `TaskExecutors.BLOCKING`?

I guess not, as I guess that guide will be retired?

sdelamo:
> Do we need to update the ExecuteOn in the MicroStream guide to `TaskExecutors.BLOCKING`?
> 
> I guess not, as I guess that guide will be retired?

we should update every guide. I will do that.

PR #1381, opened 2023-11-22
TaskExecutors.BLOCKING instead of TaskExecutors.IO

sdelamo:


guides/micronaut-oracle-autonomous-db/micronaut-oracle-autonomous-db.adoc
```diff
@@ -128,7 +128,7 @@ common:yaml-dependency.adoc[]
 
 resource:application.yml[]
 
-<1> This is optional, but it's a good idea to configure the IO pool size when using `@ExecuteOn(TaskExecutors.IO)` in controllers
+<1> This is optional, but it's a good idea to configure the IO pool size when using `@ExecuteOn(TaskExecutors.BLOCKING)` in controllers
```
timyates, line 131: This is more complex now, as I think under Java 21 it won't be taking threads from the IO pool, so I suspect this configuration will have no effect in that case...

Saying that, for Java 17 it's probably a good idea, so this should probably stay

timyates (approved):

PR #1392, opened 2023-12-11
content negotiation guide

sdelamo:
```
./gradlew clean micronautContentNegotiationBuild
```


timyates (approved):
My usual whitespace commit (I prefer ternary's with `?` and `:` at the start of the line when they are multiline?)

I also changed the test to compare the full html response...  Not sure if this is ok? Using `String.contains` makes me uneasy in tests

PR #1397, opened 2023-12-15
ReactorStreamingHttpClient and StreamedFile Guide

sdelamo:
./gradlew micronautStreamedFileAndReactorStreamingHttpClientBuild

The app works in a GraalVM Native Image but the test fails so I ignored in nativeTest

### DIFF
--- a/guides/micronaut-streamed-file-and-reactor-streaming-http-client/java/src/main/java/example/micronaut/HomeController.java
+++ b/guides/micronaut-streamed-file-and-reactor-streaming-http-client/java/src/main/java/example/micronaut/HomeController.java
@@ -2,9 +2,7 @@ package example.micronaut;
 
 import io.micronaut.context.exceptions.ConfigurationException;
 import io.micronaut.core.annotation.Nullable;
-import io.micronaut.core.util.StringUtils;
 import io.micronaut.http.HttpRequest;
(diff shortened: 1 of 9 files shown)


guides/micronaut-streamed-file-and-reactor-streaming-http-client/java/src/main/java/example/micronaut/HomeController.java
```diff
@@ -0,0 +1,97 @@
...
+    HomeController() {
+        String urlStr = "https://guides.micronaut.io/";
+        URL url;
+        try {
+            url = new URL(urlStr);
+        } catch (MalformedURLException e) {
+            throw new ConfigurationException("malformed URL" + urlStr);
+        }
+        this.reactorStreamingHttpClient = ReactorStreamingHttpClient.create(url); // <2>
+
```
yawkat:
```suggestion
```

yawkat (approved):

sdelamo:
@yawkat are there any settings you would recommend to make this example more performant? e.g. configure a separate thread pool for the streaming http client?

timyates (approved):
Fixed whitespace and unused imports

Also switched to use HexFormat instead of rolling our own

The joys of Java 17!

PR #1400, opened 2023-12-18
Remove StreamedFile mention

### LINKED ISSUE #1399: Streaming big files guide changes
> the one is a bit confusing, the title is "DOWNLOAD A BIG FILE WITH STREAMINGHTTPCLIENT" but then what is demonstrated is something different where you stream a file by reading from a URL using blocking I/O
> 
> it is unclear why blocking IO was chosen and no reference added to this can also be done with non-blocking I/O
> 
> In this case the non-blocking version would be a lot simpler since it would just be:
> 
> @Get(value="{?q}",produces="MediaType.IMAGE_PNG_TYPE")
> Publisher<ByteBuffer<?>> download(@Nullable @QueryValue String q) {
>   URI uri = q == null ? DEFAULT_URI : URI.create(q);
>         HttpRequest<?> request = HttpRequest.GET(uri);
>   return reactorStreamingHttpClient.dataStream(request);
> }
>  
> without all the complex logic to pipe reactive to blocking I/O 
> if you want this guide as it should be renamed something like "How to use blocking I/O to stream a file from a URL"

sdelamo:
Blocked via https://github.com/micronaut-projects/micronaut-core/issues/10280


guides/micronaut-streamed-file-and-reactor-streaming-http-client/java/src/main/java/example/micronaut/HomeController.java
```diff
@@ -39,44 +34,14 @@ class HomeController implements AutoCloseable {
...
-                    finallyRunnable.run();
-                })
-                .subscribe();
+    @Get // <3>
+    Flux<ByteBuffer<?>> download() {
+        HttpRequest<?> request = HttpRequest.GET(DEFAULT_URI);
+        return reactorStreamingHttpClient.dataStream(request).doOnNext(bb -> {
+            if (bb instanceof ReferenceCounted rc) {
+                rc.retain();
+            }
```
graemerocher: the content will need to be updated to explain this new code

graemerocher (changes requested):

guides/micronaut-streamed-file-and-reactor-streaming-http-client/java/src/main/java/example/micronaut/HomeController.java
sdelamo: for this code, we have a callout which says: 

> Request a stream of data where each emitted item is a ByteBuffer instance.

what do you think we should say instead?

sdelamo: ping @graemerocher

graemerocher: Something like:

```
The `dataStream` method emits instances of `ByteBuffer`. These instances are garbage collected after each chunk is emitted. In order to propagate the chunks to the response stream they need to be retained by calling the `retain()` method. The framework will call `release()` on these garbage collected chunks once each one is written to the response.
```

sdelamo: I changed it to the copy that you requested.

graemerocher (approved):

PR #1406, opened 2024-01-03
guide: Micronaut PDF OpenPDF

sdelamo:
There is no GraalVM section because nativeTest fails with: 

```
Unexpected error occurred: No awt in java.library.path
java.lang.UnsatisfiedLinkError: No awt in java.library.path
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.NativeLibrarySupport.loadLibraryRelative(NativeLibrarySupport.java:136)
        at java.base@21.0.1/java.lang.ClassLoader.loadLibrary(Class

```

There is a [related issue](https://github.com/oracle/graal/issues/2842) which seems to indicate the issue may be fixed. However, it fails for me with `17.0.8-graal` or  `21.0.1-graal` in macOS Sonoma.


timyates (approved):

guides/micronaut-openpdf/java/src/main/java/example/micronaut/PDFController.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    private void generatePDF(OutputStream outputStream) {
+        try (Document document = new Document()) {
+            PdfWriter.getInstance(document, outputStream);
+            document.open();
+            document.newPage();
+            document.add(new Paragraph("Hello World"));
+        }
```
graemerocher, line 67: more efficient would be returning `HttpResponse<Writable>`:

```java
new Writable() {
                        public void writeTo(OutputStream outputStream, @Nullable Charset charset) throws IOException {
                              generatePDF(outputStream);
                        }

                        public void writeTo(Writer out) {
                        }
                    }
```

graemerocher (changes requested):

guides/micronaut-openpdf/java/src/main/java/example/micronaut/PDFController.java
sdelamo: [I changed to use Writable](https://github.com/micronaut-projects/micronaut-guides/pull/1406/commits/4507aeeeec662aa8caece5bb18b66278d68e7b24)

graemerocher (approved):

PR #1408, opened 2024-01-05
Guide @ServerFilter @FilterMethod guide

sdelamo:


guides/micronaut-server-filter-request/java/src/test/java/example/micronaut/MemoryAppender.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+
+public class MemoryAppender extends AppenderBase<ILoggingEvent> {
+    private final List<ILoggingEvent> events = new ArrayList<>();
+
+    @Override
+    protected void append(ILoggingEvent e) {
+        synchronized (events) {
+            events.add(e);
+        }
```
timyates: Instead of this, I believe we could do:

```java
    private final List<ILoggingEvent> events = Collections.synchronizedList(new ArrayList<>());
```

Which may be better advice for people following this guide?

timyates (approved):

guides/micronaut-server-filter-request/java/src/test/java/example/micronaut/MemoryAppender.java
sdelamo: remove sync completely. Added groovy.

timyates (approved):

guides/micronaut-server-filter-request/groovy/src/main/groovy/example/micronaut/LoggingHeadersFilter.groovy
```diff
@@ -0,0 +1,50 @@
...
+@ServerFilter(MATCH_ALL_PATTERN) // <1>
+class LoggingHeadersFilter implements Ordered {
+    private static final Logger LOG = LoggerFactory.getLogger(LoggingHeadersFilter.class)
+
+    @RequestFilter // <2>
+    void filterRequest(HttpRequest<?> request) {
+        if (LOG.isTraceEnabled()) {
+            HttpHeaders headers = request.getHeaders()
+            for (String headerName : headers.names()) {
+                if (headerName.equalsIgnoreCase(HttpHeaders.AUTHORIZATION)) {
```
graemerocher: not sure we should be encouraging unsafe patterns like this, how do you know what other headers might contain sensitive values?

sdelamo: Would you prefer if I remove the part where we skip logging the Authorization header and instead log every HTTP header in the guide?

I think showing users to be careful about what they log is better than logging everything.

Please note that [in core, we skip logging some headers](https://github.com/micronaut-projects/micronaut-core/blob/4.3.x/http/src/main/java/io/micronaut/http/util/HttpHeadersUtil.java#L37-L38).  Of course, what header is sensitive is application specific.

graemerocher: perhaps you could use `HeaderUtils` since it already does some masking

sdelamo: [I have changed the Guide to use `HttpHeaderUtils`](https://github.com/micronaut-projects/micronaut-guides/pull/1408/commits/9fd2688186443555041bffafd3407392739e2e10)

graemerocher (approved):

PR #1412, opened 2024-01-08
guide: webjars stimulus

sdelamo:


guides/micronaut-webjars-stimulus/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+  "title": "Stimulus with WebJars",
+  "intro": "Learn how to load Stimulus Javascript library with WebJars",
+  "authors": ["Sergio del Amo"],
+  "categories": ["Views"],
+  "publicationDate": "2024-01-08",
+  "languages": ["java"],
+  "buildTools": ["gradle"],
```
timyates: This works as-is with Maven, is there a reason for limiting it to gradle?

sdelamo: no reason. https://github.com/micronaut-projects/micronaut-guides/pull/1412/commits/02652ec9f0e4716d65ba4fc2ae6b9676f7e308ee

graemerocher (approved):

guides/micronaut-webjars-stimulus/java/src/main/java/example/micronaut/HomeController.java
```diff
@@ -0,0 +1,17 @@
+package example.micronaut;
```
timyates, line 16: Should all these source files have the license header on them?

Or are we relying on adding the header for the generated code?

At the moment, we seem to have both 🤔

timyates: I'm going to assume we should have a license here, and I've pushed that change 👍

timyates (approved):
Pushed licenses for source code and fixed whitespace

PR #1428, opened 2024-02-01
spike: Run the guide tests in groups under a forked executor

timyates:
We have 145 guides.  This takes ages to run all the tests on, and the chances of running out of memory or hanging when they are all run on the same GitHub hosted runner is getting increasingly great.

This PR does the following:

1. Runs the test scripts in a forked jvm
    - This means we need to specify env variables in the metadata now, not at the top build level
2. Adds group takss to run a group of tests
    - In this PR we have specified 4 groups
    - If this number were in the plugin code, changing it would force a rebuild so it's in a props file in the guides folder
    - The splitting is done on a guide level, so it's not 100% fair as some guides have many apps and langs
    - However we can do this later if it proves to be necessary
    - Groups are discovered by an initial step of the build, so changing the number of groups will change the GH actions


sdelamo (commented):
This is looking great.

sdelamo (approved):

PR #1446, opened 2024-04-04
corrections to micronaut-k8s guide

dradosevic:
Some corrections to the micronaut-kubernetes guide:

- add 'yaml' to the list of Micronaut Launch features, as yaml files are used in the guide (the list of features for Micronaut CLI also contains 'yaml')
- add 'http-client' to the list of Micronaut Launch features for 'api' service (the list of features for Micronaut CLI also contains 'http-client', there is an error in the guide if 'http-client' is not included)


n0tl3ss:
The guide should use properties instead of yaml files and yaml feature needs to be removed.

sdelamo (approved):
Thanks for the contribution

PR #1453, opened 2024-04-25
Testing serialization - Spring boot vs Micronaut Framework

sdelamo:
To build the guide locally: 

`./gradlew buildingARestApiSpringBootVsMicronautTestFirstBuild`


guides/building-a-rest-api-spring-boot-vs-micronaut-test-first/micronautframeworkjacksondatabind/java/src/test/java/example/micronaut/SaasSubscriptionJsonTest.java
```diff
@@ -0,0 +1,78 @@
...
+            try (InputStream stream = optionalStream.get();
+                 Reader reader = new InputStreamReader(stream, StandardCharsets.UTF_8);
+                 BufferedReader bufferedReader = new BufferedReader(reader)) {
+                result = bufferedReader.lines().collect(Collectors.joining("\n"));
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+        return result;
+    }
```
timyates: Is this better? I suspect it doesn't matter that much...  Shame we can't more easily inject file contents into tests from test-resources

```suggestion
    private static String getResourceAsString(ResourceLoader resourceLoader, String resourceName) {
        return resourceLoader.getResourceAsStream(resourceName)
                .flatMap(stream -> {
                    try (InputStream is = stream;
                         Reader reader = new InputStreamReader(is, StandardCharsets.UTF_8);
                         BufferedReader bufferedReader = new BufferedReader(reader)) {
                        return Optional.of(bufferedReader.lines().collect(Collectors.joining("\n")));
                    } catch (IOException e) {
                        e.printStackTrace();
                    }
                    return Optional.empty();
                })
                .orElse("");
    }
```

timyates (approved):

guides/building-a-rest-api-spring-boot-vs-micronaut-test-first/micronautframeworkjacksondatabind/java/src/test/java/example/micronaut/SaasSubscriptionJsonTest.java
sdelamo: Yes, I should add a feature in Micronaut Test to ease it. Spring boot has some nice quality of live features.

graemerocher:
I'm wondering why the expected JSON needs to be in a separate file? Is there some reason for this?

sdelamo:
> I'm wondering why the expected JSON needs to be in a separate file? Is there some reason for this?

These guides are inspired by the tutorials in [Build a Rest API with Spring Boot](https://spring.academy/courses/building-a-rest-api-with-spring-boot) in [Spring Academy](https://spring.academy/). I want to compare best practices and recommendations in Spring tutorials to avoid my bias towards Micronaut Framework. They use a JSON file in those tutorials.

graemerocher:
I see, thing is that academy thing is nicely structured in terms of having learning paths etc. whilst our guides infrastructure is not like that and quite messy and difficult to find guides. Shouldn't we think about how we improve the guides to split them between tutorials, guides and having learning path pages before venturing to make it even more complicated?

sdelamo:
> I see, thing is that academy thing is nicely structured in terms of having learning paths etc. whilst our guides infrastructure is not like that and quite messy and difficult to find guides. Shouldn't we think about how we improve the guides to split them between tutorials, guides and having learning path pages before venturing to make it even more complicated?

Yes, we can introduce learning paths and courses (similar to our categories). We should spend some time creating better landing pages. 

However, those changes are independent of this PR. Can I merge this PR, or do you want to review it further?

PR #1459, opened 2024-05-03
Update to 4.4.2

sdelamo:


timyates:
The failure is with Maven and Groovy... I believe it's caused by https://github.com/micronaut-projects/micronaut-tracing/pull/559 which was in [6.5.2 of micronaut-tracing](https://github.com/micronaut-projects/micronaut-tracing/releases/tag/v6.5.2), but I'm not currently sure why

timyates:
So previously we had the core processor as an api dependency, so it was a compile dependency when we pulled in the parent

Now it's coming through as `runtime` so it doesn't work when we try to do the processing...

I guess we need fix starter to add it as a compile dependency when using jaeger, groovy and maven

guides/micronaut-microservices-distributed-tracing-zipkin/metadata.json
```diff
@@ -4,6 +4,7 @@
   "authors": ["Sergio del Amo"],
   "tags": ["distributed-tracing", "zipkin", "validation"],
   "categories": ["Distributed Tracing"],
+  "skips": [["maven", "groovy"]],
```
timyates, line 7: Added a method to skip a given buildtool and language combination

Here currently Maven and Groovy doesn't work, but instead of skipping all maven, or all groovy, we can just skip maven + groovy

PR #1490, opened 2024-10-03
Draft: extract guide generation core code

AndreaLaGrotteria:


sdelamo (changes requested):
Move the JSON Schema file to a separate PR. Place, the JSON Schema in `buildSrc/src/main/resources`. It should validate every metadata.json against the JSON Schema file. 

For example, you can validate it here: https://github.com/micronaut-projects/micronaut-guides/blob/master/buildSrc/src/main/groovy/io/micronaut/guides/GuideProjectGenerator.groovy#L84

PR #1494, opened 2024-10-07
Guides json schema

AndreaLaGrotteria:

PR #1496, opened 2024-10-09
add java classes for Json Schema

sdelamo:
This PR should remove https://github.com/micronaut-projects/micronaut-guides/blob/master/buildSrc/src/main/groovy/io/micronaut/guides/GuideMetadata.groovy and replace it with Java classes with tests which generate the same schema as the one introduced in: 
https://github.com/micronaut-projects/micronaut-guides/pull/1494


buildSrc/src/main/java/io/micronaut/guides/core/App.java
```diff
@@ -0,0 +1,54 @@
...
+import io.micronaut.core.annotation.Nullable;
+import io.micronaut.jsonschema.JsonSchema;
+import io.micronaut.serde.annotation.Serdeable;
+import io.micronaut.starter.application.ApplicationType;
+
+import java.util.List;
+
+@JsonSchema
+@Serdeable
+public record App (
```
sdelamo, line 29: Populate the description via javadoc. see: 

https://github.com/micronaut-projects/micronaut-guides/blob/master/guides/micronaut-json-schema/java/src/main/java/example/micronaut/Product.java#L26-L32

buildSrc/src/main/java/io/micronaut/guides/core/Cloud.java
```diff
@@ -1,4 +1,4 @@
-package io.micronaut.guides;
+package io.micronaut.guides.core;
```
sdelamo, line 1: Add a test which verifies the order

buildSrc/src/main/java/io/micronaut/guides/core/Guide.java
```diff
@@ -0,0 +1,96 @@
...
+        Integer minimumJavaVersion,
+
+        @JsonPropertyDescription("If the guide needs a maximum Java version, define it here")
+        Integer maximumJavaVersion,
+
+        @JsonPropertyDescription("The acronym for the cloud service provider of the guide. For example, OCI for Oracle Cloud Infrastructure")
+        Cloud cloud,
+
+        @JsonPropertyDescription("Set it to true to skip running the tests for the Gradle applications for the guide")
+        Boolean skipGradleTests,
```
sdelamo, line 78: use the primitive `boolean` instead which defaults to `false`. The default for this property.

buildSrc/src/main/java/io/micronaut/guides/core/Guide.java
```diff
@@ -0,0 +1,96 @@
...
+
+        @JsonPropertyDescription("Set it to true to skip running the tests for the Maven applications for the guide")
+        Boolean skipMavenTests,
```
sdelamo, line 82: use the primitive boolean instead which defaults to false. The default for this property.

buildSrc/src/main/java/io/micronaut/guides/core/Guide.java
```diff
@@ -0,0 +1,96 @@
...
+        TestFramework testFramework,
+
+        @JsonPropertyDescription("List of additional files with a relative path to include in the generated zip file for the guide")
+        List<String> zipIncludes,
+
+        @JsonPropertyDescription("The guide's slug. If not specified, the guides folder is used.")
+        String slug,
+
+        @JsonPropertyDescription("Whether the guide should be published, it defaults to true. You can set it to false for draft or base guides.")
+        Boolean publish,
```
sdelamo, line 107: use the primitive `boolean` instead. The default for this property should be true.

buildSrc/src/main/java/io/micronaut/guides/core/Guide.java
```diff
@@ -0,0 +1,96 @@
...
+        @JsonPropertyDescription("Defaults to null; if set, indicates directory name of the base guide to copy before copying the current one")
+        String base,
+
+        @JsonPropertyDescription("The guide's environment variables")
+        Map<String, String> env,
+
+        @JsonPropertyDescription("Applications created for the guide.")
+        App apps
+) {
+        enum Language{
```
sdelamo: Use the Language in micronaut starter dependency

buildSrc/src/main/java/io/micronaut/guides/core/Guide.java
```diff
@@ -0,0 +1,96 @@
...
+                JAVA, GROOVY, KOTLIN;
+        }
+
+        enum BuildTool{
```
sdelamo: use the BuildTool in Micronaut starter

buildSrc/src/main/java/io/micronaut/guides/core/Guide.java
```diff
@@ -0,0 +1,96 @@
...
+                MAVEN, GRADLE, GRADLE_KOTLIN;
+        }
+
+        enum TestFramework{
```
sdelamo: Use the test framework in micronaut starter dependency

buildSrc/src/test/java/io/micronaut/guides/core/AppTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@MicronautTest(startApplication = false)
+class AppTest {
```
sdelamo, line 40: the test should verify the defaults for gradle test maven tests, publish etc.

buildSrc/src/main/java/io/micronaut/guides/core/JsonAssert.java
```diff
@@ -0,0 +1,12 @@
+package io.micronaut.guides.core;
+
+import io.micronaut.guides.feature.AbstractFeature;
+import jakarta.inject.Singleton;
+
+@Singleton
+public class JsonAssert extends AbstractFeature {
+
+    public JsonAssert() {
+        super("jsonassert");
```
sdelamo: why are you adding this class in this PR?

buildSrc/src/main/java/io/micronaut/guides/core/Guide.java
AndreaLaGrotteria: but if I use primitive boolean doesn't it get automatically serialized to false if no value is specified?

AndreaLaGrotteria: what I was think of doing is keeping `Boolean publish` and then do 
```
        @Override
        public Boolean publish() {
                return publish == null ? true : publish;
        }
```

buildSrc/src/main/java/io/micronaut/guides/core/JsonAssert.java
AndreaLaGrotteria: Because I needed it for the test and the PR with the JSON guide was not merged yet when I did it. I can remove it now

PR #1508, opened 2024-10-14
One to many Guide

sdelamo:


sdelamo:
@graemerocher we are getting questions about how to model OneToMany with Micronaut Data. This is a guide we could point people to. I will publish a Youtube video explaining this material and publish it to the Micronaut Youtube Channel.

guides/micronaut-data-one-to-many/java/src/main/java/example/micronaut/PhoneEntity.java
```diff
@@ -0,0 +1,24 @@
...
+import io.micronaut.data.annotation.GeneratedValue;
+import io.micronaut.data.annotation.Id;
+import io.micronaut.data.annotation.MappedEntity;
+import io.micronaut.data.annotation.Relation;
+
+@MappedEntity("phone") // <1>
+public record PhoneEntity(
+        @Id // <2>
+        @GeneratedValue  // <3>
+        @Nullable
```
graemerocher: remove `Nullable` here as it can add confusion since from a database perspective the ID is never nullable

graemerocher (approved):

guides/micronaut-data-one-to-many/java/src/main/java/example/micronaut/PhoneEntity.java
sdelamo: https://micronaut-projects.github.io/micronaut-data/4.9.3/guide/#javaRecords

> Since records are immutable constructor arguments that are generated values need to be marked as @Nullable and you should pass null for those arguments

@dstepanov does this no longer apply?

sdelamo: @graemerocher[ I added a callout ](https://github.com/micronaut-projects/micronaut-guides/pull/1508/commits/2bb9e6078b26cee60742fcf6d95c5e612911049d)

dstepanov: I don't know why it is there; you must pass null for Micronaut / DB to generate it. There might be a problem if you have two generated values, and one is being set by recreating the record, which might trigger an error because one of the values is still null.

sdelamo: since you must pass null, @graemerocher I think it makes more sense to keep the `@Nullable` annotation. Don't you think?

PR #1509, opened 2024-10-14
Move skip logic to guides core

### LINKED ISSUE #1503: Move skip logic to guides core
https://github.com/micronaut-projects/micronaut-guides/blob/master/buildSrc/src/main/groovy/io/micronaut/guides/GuideMetadata.groovy#L80-L110

AndreaLaGrotteria:


buildSrc/src/main/java/io/micronaut/guides/core/GuideUtils.java
```diff
@@ -131,4 +134,14 @@ private static void addAllSafe(Collection target, Collection src) {
             target.addAll(src);
         }
     }
+
+    static boolean shouldSkip(Guide guide, BuildTool buildTool) {
+        if (buildTool == BuildTool.GRADLE) {
```
sdelamo: Please, use BuildTool.valuesGradle(). If skipGradleTests is true it should work for either Groovy or Kotlin Gradle DSLs.

sdelamo (changes requested):

sdelamo (changes requested):
Please, resolve the conflicts.

buildSrc/src/main/java/io/micronaut/guides/core/GuideUtils.java
AndreaLaGrotteria: done

sdelamo (approved):

PR #1536, opened 2024-10-25
Simplify Macro substitution code

sdelamo:
This PR simplifies the Macro Substitution code. 

We will need to update some of the guides macros to use Asciidoctor compatible syntax. 

e.g. We will need to replace:

```
source:TeamConfiguration[tags=teamConfigClassNoBuilder|gettersandsetters]
```

with: 

```
source::TeamConfiguration[tags=teamConfigClassNoBuilder;gettersandsetters]
```


buildSrc/src/main/java/io/micronaut/guides/core/SourceBlockMacroSubstitution.java
```diff
@@ -0,0 +1,124 @@
...
+        return str;
+    }
+
+    protected String sourceTitle(
+            String appName,
+            String condensedTarget,
+            Classpath classpath,
+            String language,
+            String packageName) {
+        return (appName.equals(MacroSubstitution.APP) ? "" : (appName + "/")) +sourceConventionFolder(classpath, language) + "/"
```
AndreaLaGrotteria:
```suggestion
        return (appName.equals(MacroSubstitution.APP) ? "" : (appName + "/")) + sourceConventionFolder(classpath, language) + "/"
```

buildSrc/src/main/java/io/micronaut/guides/core/SourceBlockMacroSubstitution.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+import java.util.Optional;
+
+import static io.micronaut.guides.core.MacroUtils.findMacroLines;
+import static io.micronaut.guides.core.asciidoc.IncludeDirective.ATTRIBUTE_LINES;
+
+abstract class SourceBlockMacroSubstitution implements MacroSubstitution {
+
+    private final LicenseLoader licenseLoader;
+    private final GuidesConfiguration guidesConfiguration;
```
AndreaLaGrotteria, line 15:
```suggestion
    private final GuidesConfiguration guidesConfiguration;
    
```

buildSrc/src/test/java/io/micronaut/guides/core/RawTestMacroSubstitutionTest.java
AndreaLaGrotteria, line 1: Are you sure the test should be modified like this? If you look at the .adoc language and build tool specific file for micronaut-security-jwt-cookie, it should be like it was before

AndreaLaGrotteria: `rawTest` should include source code from groovy, regardless of the chosen language and build

PR #1553, opened 2024-11-11
Removes metadata parsing from GuideUtils.

AndreaLaGrotteria:


AndreaLaGrotteria:
@sdelamo the removed method was only used in unitTests of buildSrc so the PR so does not affect the legacy files that generate the website

sdelamo (approved):

PR #1577, opened 2025-03-25
Add graalpy gradle plugin

tomasstupka:


guides/micronaut-graalpy-python-package/metadata.json
```diff
@@ -4,16 +4,14 @@
...
-  "buildTools": ["maven"],
+  "buildTools": ["maven", "gradle"],
   "categories": ["GraalPy"],
   "minimumJavaVersion": 21,
-  "skipMavenTests": true,
   "publicationDate": "2024-07-01",
   "apps": [
     {
       "name": "default",
-      "invisibleFeatures": ["graalpy-pygal"],
```
tomasstupka, line 15: `micronautGraalpyPythonPackageRunTestScript` failed randomly:
- the `graalpy` feature class `io.micronaut.starter.feature.graallanguages.Graalpy` in `micronaut-starter` returns an empty list from the packages method `pythonPackages()` because the launcher generated project is not supposed to have any python packages preconfigured in the `graalpy-plugin` configuration in `pom.xml` or `build.gradle` 
- the `graalpy-pygal` invisible feature class `io.micronaut.guides.feature.MicronautGraalpyPygal` in this project inherits from `Graalpy` and overrides `pythonPackages()` to return `"pygal==3.0.4"` which is the python package used`guides/micronaut-graalpy-python-package` 

seems that the order in which the features are called is not deterministic and in case, that `graalpy` feature is called second, it regenerates the `graalpy-plugin` section in `pom.xml` without any packages and the test fails. 

using only the `graalpy-pygal` resolved the issue for me

Issue #1582, opened 2025-05-20
OpenAPI Server Generation guide has broken code listing


sdelamo:
![Image](https://github.com/user-attachments/assets/2fd5e741-335a-4c9f-bc81-8d7100525db3)

PR #1599, opened 2025-10-20
Fix broken code listing in OpenAPI Server Generation guide

andreiivanou3:
Replaced the removed general `header` with the existing tags `package|import|clazz`.
Fixed the package tag that was missing its `end` marker.

Screenshot from 2025-10-20 13-50-29

#1582


sdelamo (approved):
Thanks for the contribution

PR #1798, opened 2026-05-26
Removed usage of java.util.Optional from groovy and kotlin guide

msupic:


graemerocher (approved):

guides/micronaut-data-oracle-geospatial-nearby-drivers/groovy/src/main/groovy/example/micronaut/DeliveryDispatchController.groovy
```diff
@@ -32,7 +32,7 @@ class DeliveryDispatchController {
     }
 
     @Get('/nearest-driver') // <2>
-    Optional<DriverMatch> nearestDriver(@QueryValue double longitude, @QueryValue double latitude) { // <3>
+    DriverMatch nearestDriver(@QueryValue double longitude, @QueryValue double latitude) { // <3>
```
graemerocher, line 37: not sure if this strictly equivalent, the equivalent would be to add `@Nullable` to the return type, but probably ok

